PR #5, opened 2017-07-12
Fixed wrong link path

huang-x-h:
change dynamic `import` syntax wrong link path


neoziro:
Thanks for the fix @huang-x-h!

PR #6, opened 2017-07-15
Use Promise.resolve in the events no queries are found

nylin93:
When on the server doing:

`
getLoadableState(app).then(loadableState => { ... })
`
if no queries are found, throws an error because DeferredState is returned, not a Promise.


neoziro:
Thanks, I added the test 5618b125a65e954ee0915de101377350df4532d3.

PR #20, opened 2017-11-24
forEach

stereobooster:
Some additional work required. But I wonder if you will accept this code change?

Motivation is to be able to do something like this:

```js
import { loadComponents, forEach } from "loadable-components";

// this is callback function, which will be called in the end of prerendering
// contents of returned hash (object) will be saved in window
// e.g. following code will generate
// window.__LOADABLE_COMPONENT_IDS__ = [...]
window.snapSaveState = () => {
  const componentIds = [];
  forEach(x => {
    if (x.loadingPromise) componentIds.push(x.componentId);
  });
  return {
    "__LOADABLE_COMPONENT_IDS__": componentIds
  }
}

// the rest is pretty standard React stuff
const rootElement = document.getElementById("root");
if (rootElement.hasChildNodes()) {
  loadComponents().then(() => {
    hydrate(<App />, rootElement);
  });
} else {
  render(<App />, rootElement);
}
```

Maybe there is more elegant approach. Full example here https://github.com/stereobooster/an-almost-static-stack/pull/2

### DIFF
--- a/src/componentTracker.js
+++ b/src/componentTracker.js
@@ -9,3 +9,7 @@ export const track = component => {
 }
 
 export const get = id => components[id]
+
+export const forEach = (cb) => {
+  Object.keys(components).forEach( i => cb(get(i)))
+}
--- a/src/componentTracker.test.js
+++ b/src/componentTracker.test.js
@@ -10,5 +10,9 @@ describe('componentTracker', () => {
 
     expect(componentTracker.get(id1)).toBe(Component1)
     expect(componentTracker.get(id2)).toBe(Component2)
+
+    const components = [];
+    componentTracker.forEach((x) => components.push(x))
+    expect(components).toEqual([Component1, Component2])
   })
 })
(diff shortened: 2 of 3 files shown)


neoziro:
Hello, I need more context to understand the problem you want to solve.

The code looks complicated and it leaks private implementation. I could be OK but current implement is too low level.

Can you please explain the complete use-case you are trying to solve?

stereobooster:
Yes of course. Instead of SSR I'm using technique called "snapshoting". See this [medium article](https://medium.com/superhighfives/an-almost-static-stack-6df0a2791319). The idea is simple - use real browser, crawl react application and save html files. Technique perfect for small static websites.

I want to solve "flash" of not loaded components in prerendered ("snapshoted") pages. You already solving this in SSR. This is what needed to solve this in snapshots. Ask more questions.

stereobooster:
Basically what I need - is the same function that you provide for server, but for client side. So I will be able to call it and get 

```json
{
    "__LOADABLE_COMPONENT_IDS__": [...]
  }
```

neoziro:
@stereobooster is #21 good for you?

stereobooster:
Yes totally. You are magician)

PR #21, opened 2017-11-25
feat: support snapshoting

neoziro:
Fix #20


README.md
```diff
@@ -223,6 +225,22 @@ Dynamic `import` syntax is natively supported by Webpack but not by node. That's
...
+To get it works, you have to know what components have been loaded. loadable-components exposes an API in order to make thing easier for you.
+
+```js
+import { getState } from 'loadable-components/snap'
+
+// This is callback function, which will be called in the end of prerendering
+// contents of returned hash (object) will be saved in window
+// e.g. following code will generate
+// window.__LOADABLE_COMPONENT_IDS__ = [...]
+window.snapSaveState = () => getState()
```
stereobooster: need to warn you that `snapSaveState` is specific to https://github.com/stereobooster/react-snap

stereobooster: I would say something like:

You need to instruct your snapshot solution to save state of `loadable-components` to the `window` in the end.

`getState()` will return `{__LOADABLE_COMPONENT_IDS__: [...]}`, and this should be converted to  `<script>window.__LOADABLE_COMPONENT_IDS__ = [...]</script>` in the resulting html.

For example, to do this with `react-snap` you can use following code `window.snapSaveState = () => getState()`

stereobooster:
Update react-snap documentation https://github.com/stereobooster/react-snap/pull/58 ❤️

PR #22, opened 2017-12-02
Custom components doesn't works with SSR

IcanDivideBy0:
Custom loadable components are not tracked by `componentsTracker`.
I just exposed `componentsTracker` so we can use it to track our custom components and give them an ID.
Also updated the doc :)


neoziro:
@IcanDivideBy0 thanks

Issue #10, opened 2017-07-21
Uncaught TypeError: Cannot read property 'Symbol(loadable)' of undefined


max-mykhailenko:
Problem occurs while I work with server render and dynamic client code from webpack. Seems I have difference in module numbers when HMR recompile my client code.
Server rendering works throw babel-node and don't need webpack build.
```js
/* eslint-env browser */
function loadComponents() {
  if (typeof window === 'undefined') {
    throw new Error('`loadComponents` must be called client-side: `window` is undefined');
  }

  var ids = window[_constants.COMPONENT_IDS] || [];
  return Promise.all(ids.map(function (id) {
    return componentTracker.get(id)[_constants.LOADABLE]().load(); // error here
  }));
}
```

HTML contains IDS
`<script>window.__LOADABLE_COMPONENT_IDS__ = [6553];</script>`

Thanks!

neoziro:
Hello, can you please give me the full environment? 6553 is very huge number, I don't think you have 6553 code splitted modules!

max-mykhailenko:
I have only 4 chunks. I can't show you full code because I already switch it to my own solution.

max-mykhailenko:
Can you describe some parts of business logic?
* server renders html after `getLoadableState()` which needed for `loadableState.getScriptTag()`
* `getScriptTag()` return some components IDs
* page loads, `loadComponents()` loads components with IDs from previous step. Why do we need that? I thought that `BrowserRouter` should load modules according to routing.

So, why we need `loadComponents` and `getLoadableState`

neoziro:
It is needed to prevent the page blink at load. We have to require all modules before starting to render application.

neoziro:
I think you have a problem of server-side module caching, that's why the id is 6553.

max-mykhailenko:
I render server side (in development) using babel-node without webpack. Maybe problem there. What do you think?

neoziro:
I am doing the same and I don't have any problem...

neoziro:
Have you solved your problem? I would be happy to see all the stack to understand the problem. Do you have Hangout?

max-mykhailenko:
I didn't solve it. I move to my own solution. It's more verbose but without loop over the components tree. I download component corresponding to current route.
Client and server render also without promise resolving. Several routing files allows to achieve that.

neoziro:
@max-mykhailenko not very scalable but it works ;)

grydstedt:
[...]

PR #27, opened 2018-01-11
Add support to HMR

FabioAntunes:
This is a fix for #10 

The reason why that error happens in the first place is because eveyrtime there's an HMR, the tree is recalculated, and since `currentId` and the `components`, don't get a reset on the server side, it keeps adding to those variables.

When a page is refreshed on the fronted, the tracker starts empty, so `currentId = 0` and `components= {}` creating a mismatch on the ids of the components from the backend to the frontend.

small example, if we have this routes.js:

```
// Routes.js
import loadable from 'loadable-components'

export const Home = loadable(() => import('./Home'))
export const About = loadable(() => import('./About'))
export const Contact = loadable(() => import('./Contact'))
```

**componentTracker.js** will be:
```
currentId = 3
components = {
  0: loadableFunc, // Home
  1: loadableFunc, // About
  2: loadableFunc, // Contact
}
```

if there's a HMR event on the server and frontend it will be: 

```
currentId = 6
components = {
  0: loadableFunc, // Home
  1: loadableFunc, // About
  2: loadableFunc, // Contact
  3: loadableFunc, // Home
  4: loadableFunc, // About
  5: loadableFunc, // Contact 
}
```

As long as we don't refresh the front end they will both be in sync, as soon we do a refresh, for example on the route  `/home`

server's  *componentTracker.js*:

```
currentId = 6
components = {
  0: loadableFunc, // Home
  1: loadableFunc, // About
  2: loadableFunc, // Contact
  3: loadableFunc, // Home
  4: loadableFunc, // About
  5: loadableFunc, // Contact 
}
```

it will return the id 3, because it's our current id for the **HomeComponent**:

But on the frontend our **componentTracker** will be:
```
components = {
  0: loadableFunc, // Home
  1: loadableFunc, // About
  2: loadableFunc, // Contact
}
```

And that's why we get an exception.

Now the function itself isn't the solution but it's part of it, the next step is simple adding this to our HMR part on the server and on the client:

**myServer.js**
```
import { loadableHMR } from 'loadable-components/componentTracker';
import app from './server';

if (module.hot) {
  module.hot.accept('./server', () => {
    console.log('🔁  HMR Reloading `./server`...');
    loadableHMR();
  });

  console.info('✅  Server-side HMR Enabled!');
}
```

**myClient.js**
```
import { loadableHMR } from 'loadable-components/componentTracker';

if (module.hot) {
  module.hot.accept();
  // hot.accept doesn't accept a callback, when the first arg is undefined  so we use the dispose
  module.hot.dispose(() => {
    loadableHMR();
  });
}
```

Basically what's happening in there, is whenever there's a HMR, just reset our componentTracker, so everyone will be in sync again. It may also prevent memory leaks in the future if we had zillions of HMR events along with zillions of loadable components

PS:
Do you want me to add a simple test for the function as this might decrease code coverage?

PPS:
Added tests anyway

### DIFF
--- a/src/componentTracker.js
+++ b/src/componentTracker.js
@@ -8,5 +8,6 @@ export const track = component => {
   return id
 }
 
+export const loadableHMR = () => { currentId = 0 }
 export const get = id => components[id]
 export const getAll = () => ({ ...components })
--- a/src/componentTracker.js
+++ b/src/componentTracker.js
@@ -9,8 +9,8 @@ export const track = component => {
 }
 
 export const loadableHMR = () => {
-  currentId = 0;
-  components = {};
+  currentId = 0
+  components = {}
 }
 export const get = id => components[id]
 export const getAll = () => ({ ...components })
--- a/src/componentTracker.js
+++ b/src/componentTracker.js
@@ -1,5 +1,5 @@
 let currentId = 0
-const components = {}
+let components = {}
 
 export const track = component => {
   const id = currentId
@@ -8,6 +8,9 @@ export const track = component => {
   return id
 }
 
-export const loadableHMR = () => { currentId = 0 }
+export const loadableHMR = () => {
+  currentId = 0;
+  components = {};
+}
 export const get = id => components[id]
 export const getAll = () => ({ ...components })
--- a/src/componentTracker.test.js
+++ b/src/componentTracker.test.js
@@ -16,4 +16,14 @@ describe('componentTracker', () => {
       [id2]: Component2,
     })
   })
+
+  it('should be possible to reset tracking', () => {
+    componentTracker.track(() => null)
+    componentTracker.track(() => null)
+
+    componentTracker.loadableHMR()
+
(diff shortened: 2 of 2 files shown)


neoziro:
Thanks for this PR! I would like to wait for React Hot Loader v4 to be released before merging this approach. I would like a zero-config approach to make it works out of the box with React Hot Loader v4.

FabioAntunes:
That's understandable since this is just a workaround, but at least it makes development easier until Hot loader v4 get's released. Right now it 's quite annoying to kill the server every time. 

This PR just allows to reset the counter manually, it doesn't adds any breaking changes or messes with the API internals.

Any predictions when hot loader v4 will be released?

neoziro:
In one or two weeks I hope!

neoziro:
I don't want to add code in this library to support HMR, if you want it, please use the next version of [React Hot Loader](https://github.com/gaearon/react-hot-loader/tree/next).

PR #49, opened 2018-03-02
Change name passed in to renderToString

j-hannah:
Hey guys! First of all, great library! It's the best React code splitting library I've used. I may have found something in the docs that needs a correction or clarification. I didn't see `<YourApp />` declared anywhere and using `app` works for me. If I've got that correct, here's a PR to help with make the change.

Thanks again for the great work!


neoziro:
Thanks!

PR #51, opened 2018-03-13
Change onHover -> onMouseOver

terrierscript:
I found README example mistake that `onHover` is ignored.
It's may be `onMouseOver`.

```
Warning: Unknown event handler property `onHover`. It will be ignored.
```


neoziro:
Thanks!

PR #53, opened 2018-03-21
Reload component every time when update async component code in  development mode

likun7981:
When using code splitting, additional `hot(module)` are needed because the updated components are not reloaded. Every time we need to add `hot(module)` to track the changes of asynchronous components. This makes me feel very troublesome, so every time I call the `loadable` method in `development mode`, I'm going to reload the asynchronous components once by default.  
This eliminates the need for additional `hot(module)` inside each asynchronous component.

Only one `hot(module)(App)`!

### DIFF
--- a/src/loadable.js
+++ b/src/loadable.js
@@ -90,7 +90,10 @@ function loadable(
       return <LoadingComponent {...this.props} />
     }
   }
-
+  // development mode load every time
+  if(process.env.NODE_ENV === 'development'){
+    LoadableComponent.load()
+  }
   LoadableComponent[LOADABLE] = () => LoadableComponent
 
   if (modules) {


neoziro:
Oh this look very nice, I would like to try it before but if it works, I will definitely do it!

I was looking for a solution to avoid marking each sub component with `hot`, if your solution is working this is great! 🎉

neoziro:
Ping @theKashey

theKashey:
Brilliantly elegant solution. Just wondering why I never think about such easy way to handle things.

But it shall not work - according to React-Hot-Loader tests (https://github.com/gaearon/react-hot-loader/blob/master/test/proxy/static-property.test.js#L141) it will preserve the __OLD__ value of loadingPromise.

It was made to keep the cache, people can store in static props. Actually - exactly this case.

The operation is legit only after RHL swap classes. So - just before .load - create an instance of loader.
```js
 // development mode load every time
  if(process.env.NODE_ENV === 'development'){
    React.createElement(LoadableComponent);
    LoadableComponent.load()
  }
```
Then - should work. I hope.

theKashey:
Or even better `React.createElement(LoadableComponent).type.constructor.load()`

likun7981:
Maybe I didn't use static props before, so I didn't notice this.

theKashey:
That’s not about static props, but about RHL behaviour.

neoziro:
@theKashey you are wrong.

A `load` is sufficient because the loadable component is recreated each time the file is reloaded.

neoziro:
Now added thanks!

theKashey:
Look like it's a more thicky and far more easier that I thought.
1. .load is work cos you are just calling a function
2. it does nothing except import
3. by importing a new version of file, you are re-registering top level variables.
4. thus, on next render, the old rendered component will be replaced by a new one, leading all further render to a new path.
5. Profit.

No conditions on componentWillRecieveProps, no updating for the internal promise, nothing.
__Just execute the file!__

It's brilliant! 🥇 🥇 🥇 🥇

theKashey:
But now one count not”develop” a better code spiriting expirence.
It is literally disabled.

src/loadable.js
```diff
@@ -90,7 +90,10 @@ function loadable(
       return <LoadingComponent {...this.props} />
     }
   }
-
+  // development mode load every time
+  if(process.env.NODE_ENV === 'development'){
+    LoadableComponent.load()
```
theKashey: 1. This code should be executed only of the second call of this function.
2. This code could be executed each time in node.js environment.

PR #60, opened 2018-04-05
Fix/circular

malstoun:
In some cases (when we had one error) circular structure was created in error object.

Now, if we will have one error we throw it without linking it to itself in queryErrors.


neoziro:
Thanks!

PR #65, opened 2018-04-18
feat: support new context API

neoziro:


neoziro:
@gaearon I opened an issue in React https://github.com/facebook/react/issues/12640

PR #69, opened 2018-05-03
Update README

lucasterra:
On the README, on why you don't like `react-loadable` I think you meant a 'webpack plugin', because `loadable-components` relies on a 'babel plugin' too.


neoziro:
Actually it is a Webpack + a Babel plugin.

lucasterra:
Updated.

neoziro:
@lucasterra thanks for this PR and the PR on razzle!

PR #75, opened 2018-05-08
Fix typo and improve error message

tazsingh:
`your are` -> `you are`
Also specify that the configuration is missing in the Babel config.


neoziro:
Thanks!

Issue #79, opened 2018-05-12
TypeScript support


neoziro:
As suggested in #14 it would be awesome to have a TypeScript support for Loadable Components.

This issue consist by adding Loadable Components types on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).

You can inspire by [react-loadable types](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-loadable).

theKashey:
I am 100% sure that _home_ repo is the best place for types.

neoziro:
Yeah it is a good option.

theKashey:
Then assign it to me

neoziro:
I can't but if you want to contribute you are welcome!

theKashey:
🤘🔥

PR #80, opened 2018-05-13
feat: TypeScript definitions

theKashey:
Adding types for TypeScript. implements #79
dtslint is not properly working for the case of '/server' so I had to use a bit "old" syntax for declaration.

Not adding `typecheck:ts` into CI, as long typechecking is __failing__.

dtslint was just created for DefinitelyTyped and few things are bound to DefinitelyTyped.
It asks you to use "new" syntax, dont use "declare modules" and mark all the exports as "exports". Meanwhile in this case tests could not see '/server' and there is no way to use "new" syntax.

For the old syntax it asks __not__ to use exports at all, as result - nothing will work. 💩 

So type-checking is red, type-testing is green. And type-testing is the only thing the customers needs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "repository": "git@github.com:smooth-code/loadable-components.git",
   "main": "dist/loadable-components.cjs.js",
   "jsnext:main": "dist/loadable-components.es.js",
+  "types": "types/index.d.ts",
   "module": "dist/loadable-components.es.js",
   "author": "Greg Bergé <berge.greg@gmail.com>",
   "keywords": [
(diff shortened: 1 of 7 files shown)


jbadeau:
Any timeline when this PR will be published?

theKashey:
@jbadeau - it __was__ published.

neoziro:
@jbadeau types are no longer maintained into this repository, please use https://github.com/DefinitelyTyped/DefinitelyTyped instead.

allen-munsch:
For anyone else looking for how to install the types, `@types/@loadable/components` to save you some digging its:

`npm install @types/loadable__component`

nandorojo:
And for yarn, `yarn add -D @types/loadable__component` 😊

cbdeveloper:
> For anyone else looking for how to install the types, `@types/@loadable/components` to save you some digging its:
> 
> `npm install @types/loadable__component`

This should be in the docs `README.md`. Just had to dig deep in order to find this.

I mean, those `@types` are still being used, right?

PR #83, opened 2018-05-14
chore: clone last 5 commits

DanielRuf:
By default Travis clones the last 50 commits. 5 are sufficient and faster.

See https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth


neoziro:
Thanks for the tip!

PR #87, opened 2018-05-23
On client-side init error return rejected promise rather than throw

damassi:
This swaps out an error with a rejected promise so that client UI doesn't break. Noticed when used in conjunction with [`@artsy/express-reloadable`](https://github.com/artsy/express-reloadable) that when reloads occur and module ids increment an error occurs when really there's no problem. 

(Really excellent lib -- was totally painless to setup compared to all of the others 👍)

### DIFF
--- a/src/loadComponents.js
+++ b/src/loadComponents.js
@@ -10,29 +10,26 @@ function loadState(rootState) {
       const component = componentTracker.get(state.id)
 
       if (!component) {
-        console.warn(
+        console.warn( // eslint-disable-line
           'loadable-component client modules:',
           componentTracker.getAll(),
         )
-        console.warn(
+        console.warn( // eslint-disable-line
           'loadable-component server modules:',
           window[LOADABLE_STATE],
         )
-        throw new Error(
-          `loadable-components: module "${
-            state.id
-          }" is not found, client and server modules are not sync. You are probably not using the same resolver on server and client.`,
-        )
+
+        return Promise.reject(new Error(`loadable-components: module "${
+          state.id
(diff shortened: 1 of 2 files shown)


damassi:
Ah! Updated.

damassi:
I've updated tests to cover each error branch but I can't figure out how to stub `window` in jest in order to get to that first one flagged in codecov/patch.

src/loadComponents.js
```diff
@@ -18,21 +19,23 @@ function loadState(rootState) {
...
-          `loadable-components: module "${
-            state.id
-          }" is not found, client and server modules are not sync. You are probably not using the same resolver on server and client.`,
-        )
+        const errorMessage = `loadable-components: module "${
+          state.id
+        }" is not found, client and server modules are not sync. You are probably not using the same resolver on server and client.`
+
+        console.error(errorMessage)
+        return Promise.reject(errorMessage)
```
neoziro: I think we should just do `return Promise.reject(new Error(...))`, it should be sufficient. So :

- Please remove log
- Please call `Promise.reject` with an error

neoziro:
@damassi don't worry about test coverage it is not a problem.

src/loadComponents.js
damassi: `return Promise.reject(new Error(...))` is much cleaner 👍

neoziro:
@damassi tests are broken :/

damassi:
@neoziro - looks like a CI thing -- let me kick off another commit...

neoziro:
thanks!

Issue #91, opened 2018-05-27
Just an Idea - babel-plugin - api improvement - customization


Aetherall:
First, thanks for this package ! It suits perfectly my needs :)

The only downside I see is the fact that the babel plugin does not allow customization

For exemple, I would like to have a component like this :
```js
/// /src/routing/Async.js
import loadable from 'loadable-components'

export default (importFn) => loadable(importFn, {
    loadingComponent: () => <div>My loader is too awesome</div>
})


/// /app/index.js
import { Async } from 'omni:routing' // in my project this is an alias for /src/routing/index.js (which exports the Async component)

const Counter = Async(()=>import('./counter'))

const App = () => <Counter/>

export default App

```
I edited localy the babel plugin to allow to pass an option which defaults to "loadable-components"

```json
{
    "plugins": [
        [ "loadable-components/babel", {
            "applyOnModuleName": "omni:async"
        }]
     ]
}
````

this allows me to do :

```js
/// /app/index.js
import Async from 'omni:async'

const Counter = Async(()=>import('./counter'))

export default () => <Counter/>
```
i can now alias 'omni:async' to loadable-components to get it to work but no customization possible due to the way the babel plugin actually works

my idea is to change the babel plugin to do this :

```js

loadable(() => import('./counter'))
// babel magic =>
loadable((() => {
  const fn = () => import('./counter')
  fn.modules = babelGrabbedModnames
  return fn
})())

```

so for any importFn function containing import() statements, the babel plugin will traverse the function, grab the modules used, and store them on the promise itself
this way, we can get them in the loadable function directly, and add the ability to create our own Async components without code duplication :)

neoziro:
Hi @Aetherall, first thanks for your interest!

I would like to know the original motivation of making an alias of `loadable-components`?

Also your idea is interesting but why not just giving a name option to Babel plugin?

Aetherall:
> I would like to know the original motivation of making an alias of loadable-components?

I wanted to reexport a function that has already options set (currying), so I can use custom defaults

> Also your idea is interesting but why not just giving a name option to Babel plugin?

The whole point is to make the transformation agnostic, so it's contained into the import function and allow further customizations in the future


I already achieved my goals in the fork I made,
I also added getDerivedStateFromProps support and tried and failed to add react 16.4 support 
(remove the _ of _context )

neoziro:
It concerns v2, I close it because it is outdated.

PR #92, opened 2018-05-27
Updated babel Plugin to no longer write in the second parameter + option to change the trigger module

Aetherall:
an implementation of #91


neoziro:
@Aetherall maybe the `defaultOption` is a simpler approach that solves all cases.

neoziro:
Loadable Components v3 has been released 🎉, please give a try.

👉 https://github.com/smooth-code/loadable-components

PR #95, opened 2018-05-30
Fix server-side type definitions

jiripudil:
Hello, I'm trying to use this awesome package with Typescript on the server, and I've had some difficulties. 

When I `import {getLoadableState} from 'loadable-components/server'`, The Typescript compiler complains that it "Could not find a declaration file for module 'loadable-components/server'." I've managed to work around this by explicitly setting the path mapping in tsconfig.json to `"loadable-components/*": ["node_modules/loadable-components/types"]`, but I feel that's an unnecessary step for the end user.

This PR thus extracts the server-side type definitions to a separate `server.d.ts` to match the structure of the JS modules, which seems to be [the recommended way](https://github.com/Microsoft/TypeScript/issues/8305#issuecomment-214815674).

I've also added a type definition for the `DeferredState` which wasn't exposed before.

### DIFF
--- a/types/tsconfig.json
+++ b/types/tsconfig.json
@@ -7,6 +7,11 @@
     "noImplicitThis": true,
     "strictNullChecks": true,
     "noEmit": true,
-    "strictFunctionTypes": false
+    "strictFunctionTypes": false,
+    "baseUrl": ".",
+    "paths": {
+      "loadable-components": ["."],
+      "loadable-components/server": ["../server"]
+    }
   }
 }
--- a/server.d.ts
+++ b/server.d.ts
@@ -1,0 +1,16 @@
+// TypeScript Version: 2.4
+
+import * as React from 'react';
+
+export function getLoadableState(
+  rootElement: React.ReactElement<{}>,
+  rootContext?: any,
+  fetchRoot?: boolean,
+  tree?: any,
+): Promise<DeferredState>;
+
+export interface DeferredState {
+  getScriptContent(): string;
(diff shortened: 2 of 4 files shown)


types/tsconfig.json
```diff
@@ -7,6 +7,11 @@
     "noImplicitThis": true,
     "strictNullChecks": true,
     "noEmit": true,
-    "strictFunctionTypes": false
+    "strictFunctionTypes": false,
+    "baseUrl": ".",
+    "paths": {
+      "loadable-components": ["."],
```
theKashey: This is how it could be fixed!?

jiripudil: What do you mean? The module resolution works out of the box as soon as this package is in node_modules, but within the repository, the compiler seems to need a little hint to be able to locate the module. I don't know if it's the best way, but it works :)

theKashey: I wrote the initial TS definitions, and fail to get dtslint working with server as a separate file.
You just fixed it, it the quite straight way, the way I could not figure out of dtslint documentation.
According to them - _everything_ should work even without patching tsconfig, and I will not be surprised if it will, but they wont 💩 :)

jiripudil: Submodules are apparently more complicated :) I'm not too fond of the fact that while most type definitions are in `types/`, the `server.d.ts` suddenly has to be in the package root. But it seems to be the way to go 🤷‍♂️

neoziro:
Thanks!

PR #96, opened 2018-06-01
Add a fallback component ID to `loadState`

skyuplam:
- Add a fallback component ID (i.e. component name without the index number) to `loadState` to 
  support the situation where HMR done in both server and client then refresh client browser manually.
  Such that it fixes the unmatched component state issue between server and client modules.

- Add a test case for simulating the situation

When hot reloading is done in the server side and manually refreshed the browser, the component ID in the server and client will be out of sync. For example, `./componentA-1` in server and `./componentA` in client.

In this case, `./componentA` is actually the latest version because of the manual refresh, i.e. `./componentA-1` in server since HMR is done in server side and there is no re-build/restart the server.

Thus, the index number in component ID from the server should be ignored in this situation to match the one in client.

### DIFF
--- a/src/loadComponents.js
+++ b/src/loadComponents.js
@@ -7,7 +7,10 @@ function loadState(rootState) {
 
   return Promise.all(
     rootState.children.map(state => {
-      const component = componentTracker.get(state.id)
+      const component = typeof state.id === 'string' ?
+        componentTracker.get(state.id) ||
+          componentTracker.get(state.id.substring(0, state.id.lastIndexOf('-'))) :
+        null;
 
       if (!component) {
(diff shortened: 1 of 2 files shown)


neoziro:
@skyuplam I think it could be dangerous, this issue hides a biggest problem.

I would like to think about a complete solution before taking shortcut.

skyuplam:
@neoziro Yes, I agree, and I am thinking of using hashing instead of counter for the ID.

neoziro:
Loadable Components v3 has been released 🎉, please give a try.

👉 https://github.com/smooth-code/loadable-components

PR #98, opened 2018-06-11
Add missing import to SSR part of the README.md

petetnt:
That particular sample didn't include the Route import from `react-router`


neoziro:
Thanks!

PR #104, opened 2018-06-23
Support non-ES modules

### LINKED ISSUE #103: Add support for any webpack assets
Hello! Thanks for your great lib, it greatly simplified route-based code splitting with SSR for me.

I would like to use it not just for React components, but for any modules Webpack can import. In the code below I use it to load the texts of markdown files. Whereas it works fine client-side, I get empty string when rendering server-side.

Is it hard to implement? Or maybe you can suggest a workaround?

```javascript
if (useMarkdown) {
  const pathNoExt = sourcePath.replace(/\.md$/, '');
  return {
    component: loadable(() => import(`content/${pathNoExt}.md`), {
      render: ({ Component, loading, ownProps }) => {
        if (loading || !Component) return null;
        const { default: source } = Component;
        return (
          <MarkdownArticlePage {...ownProps} source={source} />
        );
      }
    })
  };
}
```

sergeysolovev:
To fix #103

HOC `loadable` can be used for any modules that webpack can import. I’ve tested it on [my project](https://github.com/sergeysolovev/site) and it works fine for both ES modules and modules containing markdown text.

### DIFF
--- a/src/utils/resolveModuleDefault.js
+++ b/src/utils/resolveModuleDefault.js
@@ -1,5 +1,3 @@
-/* eslint-disable no-underscore-dangle */
-const resolveModuleDefault = module =>
-  module.__esModule ? module.default : module
+const resolveModuleDefault = module => module.default || module
 
 export default resolveModuleDefault
--- a/src/utils/resolveModuleDefault.test.js
+++ b/src/utils/resolveModuleDefault.test.js
@@ -1,9 +1,8 @@
 import resolveModuleDefault from './resolveModuleDefault'
 
 describe('resolveModuleDefault', () => {
-  it('should return default is ES module', () => {
+  it('should return default if present', () => {
     const module = {
-      __esModule: true,
       default: 'foo',
     }
 


src/utils/resolveModuleDefault.js
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable no-underscore-dangle */
-const resolveModuleDefault = module =>
-  module.__esModule ? module.default : module
+const resolveModuleDefault = module => module.default || module
```
theKashey: What is the real problem - `.default` for non ES module, or `not .default` for ES module?

sergeysolovev: >What is the real problem - .default for non ES module, or not .default for ES module?

`.default` for non ES module.

The problem is inconsistency between dynamic imports of webpack 4 vs. babel-plugin-dynamic-import-node. This is a known issue and currently there are 2 workarounds. Both of them are described here: https://github.com/airbnb/babel-plugin-dynamic-import-node/issues/47. 

This PR replaces one of these workarounds (deleted code) with the other one (added code). As a result, `loadable-components` will have better support for webpack 4.

theKashey: React-loadable and universal-component are both looking for `_esModule`
Async and Imported component are looking only for `.default`

@neoziro - it's your turn.

neoziro:
@sergeysolovev thanks for the fix. I think non ES modules exposing a `.default` key in their default exports are very very uncommon especially in code splitting use-cases. So I consider this change like a compatibility improvement rather than a breaking change.

sergeysolovev:
@neoziro thank you for the great lib!

PR #109, opened 2018-08-11
fix: update readme examples

bnlb:
This PR updates some code examples in the readme.

The `pMinDelay`, `delay`, and `timeout` helpers all expect a promise as the first argument, but the examples show them receiving a function that returns a promise. This updates the examples so that the `loadable` function in each case receives the function argument instead of the helpers.


neoziro:
Thanks!

PR #111, opened 2018-08-31
Update index.d.ts

ckknight:
This adds appropriate generic arguments where necessary. Without this, `React.ComponentType` defaults to `React.ComponentType<{}>`, missing out on the props of the already-specified `T`.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -24,14 +24,14 @@ export interface ComponentTracker {
 }
 
 export interface LoadableOptions<T> {
-  ErrorComponent?: React.ComponentType;
-  LoadingComponent?: React.ComponentType;
+  ErrorComponent?: React.ComponentType<T & { error: any }>;
+  LoadingComponent?: React.ComponentType<T>;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -24,14 +24,14 @@ export interface ComponentTracker {
 }
 
 export interface LoadableOptions<T> {
-  ErrorComponent?: React.ComponentType;
-  LoadingComponent?: React.ComponentType;
+  ErrorComponent?: React.ComponentType<T & { error: any }>;
+  LoadingComponent?: React.ComponentType<T>;
   render?: (options: { loading: boolean, error: boolean, ownProps: T, Component: React.ComponentType<T> }) => React.ReactElement<T>;
```
milesj: I've noticed that `error: boolean` should actually be `error: Error | null`. 

Would also be nice if render options were extracted.

```
export type LoadableRenderOptions<T = any> = {
  loading: boolean;
  error: Error | null;
  ownProps: T;
  Component: React.ComponentType<T>;
};
```

neoziro:
Please submit types in https://github.com/DefinitelyTyped/DefinitelyTyped

Issue #112, opened 2018-09-01
Include server.d.ts in npm management


highhi:
Thank you for a nice package!

`server.d.ts` does not seem to be managed by npm.
Is it possible to modify `.npmignore` and manage server.d.ts with npm?

neoziro:
Fixed in #113

PR #113, opened 2018-09-03
Add server.d.ts to .npmignore

lucasterra:
#95 broke TS Definitions for `loadable-components/server`. That's because `server.d.ts` is not being exported on npm publish.

This fixes this #112


neoziro:
Good catch, thanks!

PR #117, opened 2018-09-19
Add note about Promise returned by load

sjsyrek:
I discovered that the `load()` function returns a Promise, so if you don't handle the error case, you get noise in the console. This isn't mentioned in the documentation, so I added a note about it.


neoziro:
Thanks!

PR #129, opened 2018-10-30
Babel plugin: prefer import than a variable name

### LINKED ISSUE #128: Server side rendering have no effect in 3.0.x
Hi!
I have writen a ssr demo using `2.2.3`,[here](https://github.com/code-mcx/react-ssr/tree/webpack3.x) is my demo.
Now i want to upgrade `loadable-components` from `2.2.3` to `3.0.x`,only a few things have been done

1. Upgrade babel to 7.x.
2. Replace all `import Loadable from "loadable-components";` with `import Loadable from "@loadable/component";`.
3. Replace babel plugin `loadable-components/babel` with `"@loadable/babel-plugin"`.
4. Using dynamic mode in `src/server.js` replace old api

.

`server.js`
```js
const ReactDOMServer = require("react-dom/server");
const { LoadableState } = require("@loadable/server");
// ...

let component = createApp(context, req.url, store);

const loadableState = new LoadableState();
const loadableComponent = loadableState.collectChunks(component);
// ...

let html = ReactDOMServer.renderToString(loadableComponent);

let htmlStr = template
    // some code
    .replace("<!--react-ssr-outlet-->", 
    `<div id='app'>${html}</div>\n${loadableState.getScriptTags()}`);
```

After call `ReactDOMServer.renderToString()`,i call `loadableState.getScriptTags()`,but return `<script>window.__LOADABLE_STATE__ = [];</script>`

Thanks

bertho-zero:
Should fix #128

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -127,8 +127,8 @@ loadable({
 `;
 
 exports[`plugin load should support naming 1`] = `
-"import load from '@loadable/component';
-load({
+"import Loadable from '@loadable/component';
+Loadable({
   chunkName() {
     return \\"moment\\";
   },
@@ -167,8 +167,8 @@ load({
 `;
 
 exports[`plugin load.lib should support naming too 1`] = `
-"import load from '@loadable/component';
-load.lib({
+"import Loadable from '@loadable/component';
+Loadable.lib({
   chunkName() {
     return \\"moment\\";
   },
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -1,7 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`plugin aggressive import should work with destructuration 1`] = `
-"loadable({
+"import loadable from '@loadable/component';
+loadable({
   chunkName({
     foo
   }) {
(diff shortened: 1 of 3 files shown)


neoziro:
Hello @bertho-zero, thank for you work, but I prefer to rely on using explicitely `loadable`. Then you could choose to use `import loadable from 'something'` and it will work. It is the same strategy used by styled-components and emotion, so let's stick to it.

bertho-zero:
styled-components uses the require, not a variable name that can be used anywhere. (https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/visitors/assignStyledRequired.js#L7)

emotion uses the import, as I did here. (https://github.com/emotion-js/emotion/blob/master/packages/babel-plugin-emotion/src/index.js#L65)

bertho-zero:
`import loadable from 'something'`

It will not be used like that, when is it useful to use the loadable-components babel plugin with another package?
But we may want to name his variable as we want and that's normal.

bertho-zero:
I can add an option to choose the name of the package to use, is it a good compromise?

neoziro:
I prefer to keep it simple for now, I don't understand why naming it "loadable" is a problem?

bertho-zero:
The babel plugin may be easier to code but it does not reflect the expected use of a babel plugin.

Why go through the AST if it is to do the equivalent of a `.replace`?

This is a plugin that should apply to imports of `@loadable/component` but instead it reserves a function name. I do not understand logic at all.

neoziro:
I want to ensure convention, by doing so you can import loadable from a pre-configured wrapper.

```
import rawLoadable from '@loadable/component'

export const loadable = func => rawLoadable(func, { fallback: 'Loading...' })
```

And it will work, you can't with your method. I think this is why `styled-components` and `emotion` are using this strategy too.

PR #133, opened 2018-11-01
add missing imports

salzhrani:
## Summary

Fix documentation by adding some missing imports

## Test plan

Not sure if necessary


neoziro:
Thanks!

PR #143, opened 2018-11-13
fix ref for load lib

philibea:
## Summary

Load a lib with a ref not working actually

## Test plan

Look en example code in `examples/client-side/src/index.js`


neoziro:
See #144

PR #155, opened 2018-11-17
Exclude http and https from regex. Should fix #153

### LINKED ISSUE #153: Breaks when Webpack's output.publicPath is a URL
When `output.publicPath` in webpack is a URL, as is the case when using a CDN, the returned script tags from `extractor.getScriptTags` have broken src URLs with a slash being removed from the protocol.

If the below is set in webpack.
```js
webpackConfig = {
  //...
  output: {
    publicPath: 'https://cdn.example.com/assets/',
  }
};
```

then `extractor.getScriptTags()` will return.
```html
<script>window.__LOADABLE_REQUIRED_CHUNKS__ = ["loadable-example"];</script>
<script async data-chunk="loadable-example" src="https:/cdn.example.com/assets/loadable-example.js"></script>
<script async data-chunk="main" src="https:/cdn.example.com/assets/main.js"></script>
```

patrickhaug:
The regex now excludes leading http (https). There was a failing test before. Should I look after that one?

### DIFF
--- a/packages/server/src/util.js
+++ b/packages/server/src/util.js
@@ -9,4 +9,5 @@ export const smartRequire = modulePath => {
   return require(modulePath)
 }
 
-export const joinURLPath = (...paths) => paths.join('/').replace(/\/\//g, '/')
+export const joinURLPath = (...paths) =>
+  paths.join('/').replace(/(?< !http :| https: ) \/\//g, '/')


patrickhaug:
Yeah, saw that too after pushing... but that failing test was something different? 😕

neoziro:
@patrickhaug thanks, can you add a test? 

To do it create a new file `util.test.js` :

```js
describe('util', () => {
  describe('#joinURLPath', () => {
    it('should join paths', () => {
      expect(joinURLPath('a', 'b')).toBe('a/b')
    })
  })
})
```

Please add the one that was causing problem, thanks!

willdurand:
Hi, I think this still fails if the `publicPath` is protocol-free (e.g., `//127.0.0.1:3001`).

patrickhaug:
@willdurand yup, this will fail. I opened a [PR](https://github.com/smooth-code/loadable-components/pull/163)

PR #161, opened 2018-11-21
add writeToDisk option to webpack plugin

dlebedynskyi:
## Summary

Adds option to emit file on build  [158](https://github.com/smooth-code/loadable-components/issues/158) by adding writeToDisk (default to false) to plugin options 

## Test plan

Test required app that runs 2 instances of webpack - client and server.  
Client webpack instance need to generate file that is later used by server instance. 
Razzle app maybe be a good candidate for that test

### Default 
```
// add plugin instance in webpack for client config
 plugins: [new LoadablePlugin()]
```

### with emit file 
```
// 👇 generate absolute path to target file 
const filename = path.resolve(root, 'dist/loadable-stats.json'); 

// add plugin instance in webpack for client config
 plugins: [new LoadablePlugin({filename:  writeToDisk: true})]

```

### DIFF
--- a/packages/webpack-plugin/README.md
+++ b/packages/webpack-plugin/README.md
@@ -20,6 +20,24 @@ module.exports = {
 }
 ```
 
+## API
+
+### LoadablePlugin
+
+Create a webpack loadable plugin.
+
+| Arguments             | Description                                       |
+| --------------------- | ------------------------------------------------- |
+| `options`             | Optional options                                  |
+| `options.filename`    | Stats filename (default to `loadable-stats.json`) |
+| `options.writeToDisk` | Always write assets to disk (default to `false`)  |
+
+```js
+new LoadablePlugin({ filename: 'stats.json', writeToDisk: true })
+```
+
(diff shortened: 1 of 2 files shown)


neoziro:
OK for me, @marcneander can you confirm that it work for your razzle example?

marcneander:
It works great 👍 I created my razzle example in his branch. Would have loved to have added it to this PR but I don't think it's possible so I created a new PR.

Let's get this merged and published. We're already running this code in our platform :)

neoziro:
I can't merge it right now, documentation is not up to date. @dlebedynskyi can you update readme?

willdurand:
> // add plugin instance in webpack for client config
> plugins: [new LoadablePlugin({filename: writeToDisk: true})]

I suppose, there is a comma after `filename`, no?

FWIW, I currently use this plugin to (I guess) do the same thing: https://github.com/gajus/write-file-webpack-plugin.

dlebedynskyi:
> I suppose, there is a comma after filename, no?
@willdurand my bad. 

Thanks for merge and coma fix @neoziro.

PR #163, opened 2018-11-22
Support protocoll free public path in joinURLPath Method

patrickhaug:
This is the fix for [missing feature in last PR](https://github.com/smooth-code/loadable-components/pull/155#issuecomment-440728651)


neoziro:
Thanks!

PR #167, opened 2018-11-26
Razzle example server statsFile fix

radum:
The Razzle example loads the JSON as is instead of passing the path to ChunkExtractor via statsFile.

This will cause an error.


neoziro:
Hello @radum, thanks for your fix, can you fix tests please?

PR #170, opened 2018-11-27
fix(webpack-plugin): fix TypeError when set writeToDisk true

code-mcx:
This is fix for [https://github.com/smooth-code/loadable-components/issues/168](https://github.com/smooth-code/loadable-components/issues/168).


neoziro:
Thanks @code-mcx !

PR #176, opened 2018-12-07
fix install command in Server Side Rendering

leejh3224:
Found a typo in `Server Side Rendering` part of website.

Added missing `install` and `add` command in installation codes.


neoziro:
Thanks!

PR #187, opened 2018-12-14
fix issue with @loadable/webpack-plugin not working in production mode and SSR

tadeuszwojcik:
This is my stab at https://github.com/smooth-code/loadable-components/issues/179 .

I've added `path` option to the webpack plugin, that allows specifying where loadable-stats are written to when `writeToDisk` option is set to true. This way webpack assets compiler always use `filename` and not absolute path when adding loadable stats to the assets. If `path` is not set by default loadable-stats are written to `output.path` set in webpack config.

I also fixed issue I had with `eval('module.require')` trick as it was throwing errors (eval is not a function). I've replaced it with `__non_webpack_require__` and that seems to be working fine.

I've tested it with razzle example in the repo and it works fine both in dev and production mode (after yarn build command).

Closes https://github.com/smooth-code/loadable-components/issues/179

### DIFF
--- a/packages/server/src/util.js
+++ b/packages/server/src/util.js
@@ -5,10 +5,16 @@ export const smartRequire = modulePath => {
     clearModuleCache(modulePath)
   }
 
-  // Use eval to prevent Webpack from compiling it
+  // Use __non_webpack_require__ to prevent Webpack from compiling it
   // when the server-side code is compiled with Webpack
-  // eslint-disable-next-line no-eval
-  return eval('module.require')(modulePath)
+  // eslint-disable-next-line camelcase
+  if (typeof __non_webpack_require__ !== 'undefined') {
+    // eslint-disable-next-line no-undef
+    return __non_webpack_require__(modulePath)
+  }
+
+  // eslint-disable-next-line global-require, import/no-dynamic-require
+  return require(modulePath)
 }
 
 export const joinURLPath = (...paths) => {
--- a/examples/razzle/razzle.config.js
+++ b/examples/razzle/razzle.config.js
@@ -12,7 +12,10 @@ module.exports = {
 
       appConfig.plugins = [
         ...appConfig.plugins,
-        new LoadableWebpackPlugin({ writeToDisk: { filename } }),
+        new LoadableWebpackPlugin({
+          outputAsset: false,
+          writeToDisk: { filename },
+        }),
       ]
     }
 
--- a/examples/razzle/razzle.config.js
+++ b/examples/razzle/razzle.config.js
@@ -8,11 +8,11 @@ module.exports = {
     const appConfig = Object.assign({}, config)
 
     if (target === 'web') {
-      const buildPath = path.resolve(__dirname, 'build')
+      const filename = path.resolve(__dirname, 'build')
 
       appConfig.plugins = [
         ...appConfig.plugins,
-        new LoadableWebpackPlugin({ writeToDisk: true, path: buildPath }),
+        new LoadableWebpackPlugin({ writeToDisk: { filename } }),
       ]
     }
 
--- a/examples/razzle/razzle.config.js
+++ b/examples/razzle/razzle.config.js
@@ -8,11 +8,11 @@ module.exports = {
     const appConfig = Object.assign({}, config)
 
     if (target === 'web') {
-      const filename = path.resolve(__dirname, 'build/loadable-stats.json')
+      const buildPath = path.resolve(__dirname, 'build')
 
       appConfig.plugins = [
         ...appConfig.plugins,
-        new LoadableWebpackPlugin({ writeToDisk: true, filename }),
+        new LoadableWebpackPlugin({ writeToDisk: true, path: buildPath }),
       ]
     }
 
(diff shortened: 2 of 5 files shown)


packages/server/src/util.js
```diff
@@ -5,10 +5,16 @@ export const smartRequire = modulePath => {
...
-  // eslint-disable-next-line no-eval
-  return eval('module.require')(modulePath)
+  // eslint-disable-next-line camelcase
+  if (typeof __non_webpack_require__ !== 'undefined') {
+    // eslint-disable-next-line no-undef
+    return __non_webpack_require__(modulePath)
+  }
+
+  // eslint-disable-next-line global-require, import/no-dynamic-require
+  return require(modulePath)
```
neoziro: Are you sure that it will not cause any error or warning when built using Webpack?

tadeuszwojcik: Yes, tried that with razzle that uses webpack for the server bundle as well

tadeuszwojcik:
@neoziro thanks for taking the time and looking into this, I've the path in the same way that https://github.com/ztoben/assets-webpack-plugin/blob/master/readme.md uses, happy to continue this conversation after Christmas as I'm on holidays right now. Please don't close/rejected it before then, I'll try to put more details about rationale of this pull request and why it's essential for razzle and razzle like setups! Thanks!

tutok:
@neoziro What is the status of this issue? Is there anything that you would like to change and/or you are not sure about? Maybe I can help somehow? 

From my perspective changes done by @tadeuszwojcik are ok. Actually, many webpack loaders/plugins follow that pattern were output path can be overridden, e.g. [file-loader](https://github.com/webpack-contrib/file-loader#outputpath)

neoziro:
@tutok my concern is that is an absolute output path is specified it will be written in two different folders: https://github.com/smooth-code/loadable-components/pull/187/files#diff-567a0d765f93649fdd2c856a649ba64bR29

I think it is not good. This is the only blocking point.

dlebedynskyi:
@tadeuszwojcik @neoziro  
When I wrote original version it was expected that when using writeToDisk, filename must be an absolute path. 
This was done for 
 - webpack dev server to actually write file on file system. 
 - most cases you don't want to output assets file at same level as you browser assets. kinda extra security level. 

currently we are using webpack dev server for client bundle and output is 
 ```
build/
 - assets.json 
 - /dist/ 
   - client.bundle.js
```
where asset.json absolute build path. 

we can check [this logic](https://github.com/themgoncalves/react-loadable-ssr-addon/blob/master/source/ReactLoadableSSRAddon.js#L126)  that I used as reference back then.

**Comment by neoziro** [comment]
With the actual code, stats will be written two times if `writeToDisk` is set to true. It is not good for me.  I think we have to change the API:

```js
new LoadablePlugin({ writeToDisk: { filename: '/your/custom/path/loadable-stats.json' } })
```

This way it is clear that the asset will be written to disk at a specific location. We can also add a config option `outputAsset` that will control if a new asset is added or not.

What do you think?

tutok:
@neoziro you mean `writeToDisk` could be `{filename: string}` or `boolean` right? for me this is ok.

btw. is even `loadable-stats.json` an webpack asset? Maybe this file should not be considered as webpack asset as it contains rather metadata about them. Moreover `loadable-stats.json` is not event used by "client" build, it is expecially created for "server" build for SSR.

neoziro:
@tutok yeah this is what I mean for `writeToDisk` option. I understand your point but I think it is simpler by default to emit it from webpack assets. Also in a specific configuration people could want it to not be written on the disk (SSR).

tadeuszwojcik:
@neoziro I've pushed the change with `writeToDisk {filename: string}` change, hopefully it will make webpack-plugin API more self explaining.

neoziro:
I am sorry, I would like you to make one last change to make it easy for users:

- `writeToDisk` can be set to `true`, `false` or `{ filename: "..." }`, it is `false` by default
- `false` will not write to disk (default value)
- `true` will write to disk using `filename` (exactly like today)
- `{ filename: "..." }` will write to disk to a custom location

Can you also add a `outputAsset` option?

- `outputAsset` can be set to `true`
- it is `true` by default
- `true` will append stats into assets (like today)
- `false` will not append stats into assets

This way we could do:

```js
new LoadablePlugin({ outputAsset: false, writeToDisk: { filename: 'my/custom/stats.json' })
```

I think the API is great and can handle all cases.

neoziro:
OK I see what is your problem, for now `globalStyle` accepts a `theme` property: 

```js
const GlobalStyle = createGlobalStyle`${globalStyle(customTheme)}`;
```

It will not be the case any more with next version (supporting emotion 10), but for now it still requires (even if you don't use emotion).

Thanks for your repro it helped a lot!

dlebedynskyi:
@neoziro  I don't think you last comment was to this issue. Probably should not close this ticket either

neoziro:
Oups sorry!

neoziro:
@tadeuszwojcik any update?

neoziro:
Thanks for your work!

PR #190, opened 2018-12-19
Inline CSS options

bigwillch:
Hi. We have a requirement to render our css inline in certain situations, and so I had a crack at implementing it with the loadable library. Please do let me know what you think about the approach.

## Summary

In order to improve page speed score, or to render AMP SSR pages alongside the main app, it can be a requirement to render css inline.

## Test plan

This has been added as an option to the existing `chunkExtractor.getStyleTags()` and `chunkExtractor.getStyleElements()` functions via an `inline` boolean parameter. 

Additionally a `chunkExtractor.getCssString()` function has been added to allow developers further control over how to render css (an example use requirement being passing the css string to a style tag using AMP's `amp-custom` attribute)

Appropriate tests have been added to `ChunkExtractor.test.js`

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -35,7 +35,10 @@ function assetToScriptElement(asset) {
 function assetToStyleString(asset) {
   return new Promise((resolve, reject) => {
     fs.readFile(asset.path, 'utf8', (err, data) => {
-      if (err) reject(err)
+      if (err) {
+        reject(err);
+        return;
+      }
       resolve(data);
     })
   })
@@ -50,7 +53,10 @@ function assetToStyleTag(asset) {
 function assetToStyleTagInline(asset) {
   return new Promise((resolve, reject) => {
     fs.readFile(asset.path, 'utf8', (err, data) => {
-      if (err) reject(err)
+      if (err) {
+        reject(err);
+        return;
+      }
       resolve(
         `<style data-chunk="${asset.chunk}">
         ${data}
@@ -75,7 +81,10 @@ function assetToStyleElement(asset) {
 function assetToStyleElementInline(asset) {
   return new Promise((resolve, reject) => {
     fs.readFile(asset.path, 'utf8', (err, data) => {
-      if (err) reject(err)
+      if (err) {
+        reject(err);
+        return;
+      }
       resolve(
         <style
           key={asset.url}
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -1,5 +1,6 @@
 /* eslint-disable react/no-danger */
 import path from 'path'
+import fs from 'fs'
 import _ from 'lodash'
 import React from 'react'
 import { invariant, LOADABLE_REQUIRED_CHUNKS_KEY } from './sharedInternals'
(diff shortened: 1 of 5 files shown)


neoziro:
Hello @bigwillch, thanks for your work, you made a great job! I have only one problem: using `fs.readFileSync` is not good in production, it blocks thread. So we have to use the asynchronous method instead. But if we use it: `getStyleTags` and `getStyleElements` would become asynchronous. So my suggestion is to create custom methods for your use-case: `getInlineStyleTags`, `getInlineStyleElements`, `getCSSString`.

Is it possible?

bigwillch:
Thanks for the feedback! I'll have a a crack at creating new async methods as suggested

neoziro:
Thanks, let me know when it is done!

bigwillch:
@neoziro just a question - would you be opposed to the package providing both sync and async methods so that the developer can choose which method to use in the app? 

I'm just thinking that the tradeoff between performance and having to wrap the html output in promises may be something developers would like to be able to choose themselves

neoziro:
@bigwillch yes we could expose the two methods, just add `getSyncInlineStyleElements` if you want. It costs nearly nothing and it could be useful for users.

neoziro:
I am not OK to return a different type (especially a promise) when a custom option is passed to the function. I think it is not clear for the user, it can be confusing and it is not type friendly.

Can you please create a different method for each one?

bigwillch:
@neoziro no problem :) Have made them separate methods as requested

MartinGian:
I'm very interested in this functionality :) I can help testing it on my project let you know if I have any issues.

packages/server/src/ChunkExtractor.js
```diff
@@ -31,12 +32,35 @@ function assetToScriptElement(asset) {
...
   return `<link data-chunk="${asset.chunk}" rel="stylesheet" href="${
     asset.url
   }">`
 }
 
+function assetToStyleTagInline(asset) {
+  return new Promise((resolve, reject) => {
+    fs.readFile(asset.path, 'utf8', (err, data) => {
+      if (err) reject(err)
+      resolve(
```
neoziro: I think you are missing a else or an early return.

packages/server/src/ChunkExtractor.js
```diff
@@ -48,6 +72,21 @@ function assetToStyleElement(asset) {
   )
 }
 
+function assetToStyleElementInline(asset) {
+  return new Promise((resolve, reject) => {
+    fs.readFile(asset.path, 'utf8', (err, data) => {
+      if (err) reject(err)
+      resolve(
```
neoziro: Same here.

neoziro:
@bigwillch can you please fix the resolve / reject and I will merge it! Thanks!

bigwillch:
@neoziro thanks for flagging! sloppy of me :) Hope it's all ok now 🙏

neoziro:
Thanks for your work @bigwillch!

bigwillch:
Great! Thanks for approving

bigwillch:
@neoziro has v5.3.0 been published to npm yet? I'm not seeing it when trying to upgrade

neoziro:
https://www.npmjs.com/package/@loadable/server

bigwillch:
sorry, I'm dumb. Was trying to upgrade loadable/component instead. doh

nesbtesh:
I am going through the code and I can’t find where do you declare which style elements you want inline? I dont want to include the entire apps css inline, just the part above the fold.

theKashey:
@nesbtesh - there is no such thing as "above the fold", without prerendering your application into some virtual window first - then you'll know what it above the fold.

You are probably looking for [used-styles](https://github.com/theKashey/used-styles#stream-rendering) in stream mode, where styles are inlined before they used, no matter of the "fold".

nesbtesh:
@theKashey that worked great, thanks, time to interactivity improved to 1s

PR #198, opened 2019-01-08
Update yarn.lock for server-side-rendering example

overlookmotel:
## Summary

The `server-side-rendering` example `yarn.lock` file references the evil [flatmap-stream](https://www.npmjs.com/package/flatmap-stream) package v0.1.1 (see https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident).

Consequently `yarn install` fails.

This PR regenerates the `yarn.lock` file to avoid this problem.

## Test plan

n/a

### DIFF
--- a/examples/server-side-rendering/yarn.lock
+++ b/examples/server-side-rendering/yarn.lock
@@ -3,9 +3,9 @@
 
 
 "@babel/cli@^7.1.2":
-  version "7.1.2"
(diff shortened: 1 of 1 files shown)


overlookmotel:
OK great. Thanks. I don't use yarn myself, so not a surprise I got it wrong.

PR #210, opened 2019-01-22
Allow to specify extra attribute in getScriptTags

nchaulet:
## Summary

When you using loadable components server side
I want to be able to provide extra attributes to generated script tags, especially a `nonce` attribute used for `Content-Security-Policy`.

## Test plan

I added a small test on how you can use this new extraProps.

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -30,9 +30,9 @@ function assetToScriptTag(asset, extraProps) {
   }"${extraPropsToString(extraProps)}></script>`
 }
 
-function assetToScriptElement(asset) {
+function assetToScriptElement(asset, extraProps) {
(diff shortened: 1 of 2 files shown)


neoziro:
Hello @nchaulet, thanks for your PR. I think it is a good idea! But I would like to have it on all methods available. Having it on `getScriptTags` and not on `getScriptElements` is pretty confusing. Same on style methods. Could you generalize it?

nchaulet:
Sure I am going to add it

neoziro:
Thanks, one last thing, can you update the documentation? Also should we add it to `getStyleXXX` methods?

nchaulet:
Yes I am going to update the doc, and add it to `getStyle*`

neoziro:
Thanks!

PR #212, opened 2019-01-24
fix: Add extra props option for links

mattsjones:
## Summary
Add extra prop options for `getLinkTags` and `getLinkElements`.

The latest version added extra prop options for all the get tag functions. It seems as though links were missed in this even though they were documented as having the option.

## Test plan

Added snapshot tests for the new options. 

Note: I ran prettier over the project but it seemed to change some extra lines. Not sure if it was upto date previously.

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -21,7 +21,10 @@ function getAssets(chunks, getAsset) {
 }
 
 function extraPropsToString(extraProps) {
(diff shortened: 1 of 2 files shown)


mattsjones:
Hey @neoziro. Not sure what you'd like updated in the docs? As I mentioned in the PR description the feature was already documented in the previous release but missed out on the actual implementation.

neoziro:
OK sorry, so I merge it! Thanks!

mattsjones:
Any chance we could release this? Would be really appreciated.

PR #221, opened 2019-02-04
feat(babel-plugin): transform code annotated with magic comment

tagoro9:
## Summary

This PR updates the babel plugin to also transform the imports inside code annotated with the `/* #__LOADABLE__ */` comment as specified in https://github.com/smooth-code/loadable-components/issues/192.

Right now the plugin is able to transform these 2 cases:

```JS
/* #__LOADABLE__ */
const loader = () => import('moment');
```

```JS
export default {
  /* #__LOADABLE__ */
  loader: () => import('moment')
}
```

Should it also consider any other case?

This is the first time I work on a babel plugin, so there may be some things to be improved in my code.

Docs should also be updated, but I'm not quite sure where it would be the best place to include this new functionality.

## Test plan

I added a few test cases for the new code and snapshots seem to be ok (compared with the other test cases).

Thanks!

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -1,5 +1,85 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`plugin Magic comment should transpile object properties 1`] = `
+"export default {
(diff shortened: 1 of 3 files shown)


neoziro:
@tagoro9 thanks for your work! See #222 for follow up.

PR #222, opened 2019-02-05
feat(babel-plugin): transform code annotated with magic comment

### LINKED ISSUE #192: Parse functions annotated
## 🚀 Feature Proposal

Actually Babel plugin relies on `loadable`, it would be great to rely on a comment too. See #188.

## Motivation

See #188.

## Example

```js
/* #__LOADABLE__ */
const loader = () => import('./MyComponent');
```

```js
var loader = function () {
  return {
      chunkName: function chunkName() {
        return "my-component";
      },
      isReady: function isReady(props) {
        if (typeof __webpack_modules__ !== 'undefined') {
          return !!__webpack_modules__[this.resolve(props)];
        }

        return false;
      },
      requireAsync: function requireAsync() {
        return import(
        /* webpackChunkName: "my-component" */
        './MyComponent');
      },
      requireSync: function requireSync(props) {
        var id = this.resolve(props);

        if (typeof __webpack_require__ !== 'undefined') {
          return __webpack_require__(id);
        }

        return eval('module.require')(id);
      },
      resolve: function resolve() {
        if (require.resolveWeak) {
          return require.resolveWeak("./MyComponent");
        }

        return require('path').resolve(__dirname, "./MyComponent");
      }
    }
```

neoziro:
Transform imports inside code annotated with `/* #__LOADABLE__ * /`.

Fixes #192


tagoro9:
Thanks so much!

PR #224, opened 2019-02-05
feat(component): add preload method

### LINKED ISSUE #196: Consider putting back prefetch function.
## Motivation

Before, I could control manually when to download a chunk / component.
This allowed me to have a fast app boot / login screen. Then I could tell manually to fetch the "Private" section. This would result in really smooth UX. Switching to Private section would be instant.

In the breaking change, we are told to use "Webpack magic comments." But Webpack does not allow to do the above. At least, I could not find it.

We can just set `/* webpackPrefetch: true */`. We are left with the "good will of the browser" and in my case it's just a step back as the switch to the Private section now is not downloaded by the browser in time and results in worst UX.

## Pitch

Unless there is an alternative that I don't know about, I'd like to see this feature back until Webpack is able to let us load manually a chunk / component, since it allows superior performance in some cases.

gregberge:
Closes #196

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -163,6 +163,13 @@ function createLoadable({ resolve = identity, render, onLoad }) {
       <EnhancedInnerLoadable forwardedRef={ref} {...props} />
     ))
 
+    Loadable.preload = props => {
+      if (typeof window === 'undefined') {
+        throw new Error('`preload` cannot be called server-side')
+      }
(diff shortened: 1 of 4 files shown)


bertho-zero:
@neoziro How is it possible to wait for the end of the preload? I would like to use the promise.

bertho-zero:
It does not make much sense that `preload` is chainable, can we return the promise ?

neoziro:
@bertho-zero see #226

PR #228, opened 2019-02-08
Fix usage with commonjs modules

bertho-zero:


theKashey:
That would be a breaking change.
The code you have written is true only for babel-compiled scripts, while TypeScript has a different "default" behaviour. 

First of all - what problem you are trying to solve?

bertho-zero:
I thought it was a universal solution.
For another configuration problem, I had a module that returned `undefined`, and I had a component `{default: undefined, __esModule: true, ...}` in React.

theKashey:
What about picking `default` when you sure you could do it, and using the old behavior in any other case?
```js
loadedModule.__esModule ? loadedModule.default :  (loadedModule.default || loadedModule)
```

theKashey:
👍

neoziro:
Thanks!

PR #229, opened 2019-02-08
Allow query-param cache busting in chunk names

jordanburk:
We use query params for cache-busting (eg, `chunkFilename: '[name].js?[chunkhash]'`), but `extname()` can't properly get the extension when there's a param attached. This change would allow the use of query params.

### DIFF
--- a/packages/server/__fixtures__/stats.json
+++ b/packages/server/__fixtures__/stats.json
@@ -13,6 +13,10 @@
     "letters-B": "letters-B.js",
     "letters-C": "letters-C.js",
     "letters-D": "letters-D.js",
+    "letters-E": [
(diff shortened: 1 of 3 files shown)


neoziro:
Hello @jordanburk, thanks, can you add a test please?

jordanburk:
Hi @neoziro , sorry about that! I have added tests as requested. Please let me know if they are sufficient, thank you ^_^

neoziro:
Yes all good for me thanks!

jordanburk:
Hi @neoziro, thank you for accepting this. Any estimate on when it will be released? Hoping to get this in use on production soon.
Thanks for your time!

neoziro:
Released!

PR #237, opened 2019-02-13
Update docs to correct syntax for prefetching example

bitttttten:
I think the docs are incorrect, the extra arrow function is not necessary and it's missing `import`.


neoziro:
Thanks!

PR #239, opened 2019-02-15
Adding basic support for .mjs file format.

ashuanindian:
As almost all major browsers already implemented es6 modules and pushing .mjs file format for the es6 modules to keep it consistent with nodejs environment. 
High performance website are considering to send different code to modern and legacy browsers.

This pull request will only help getting script tags with correct filename for both .js and .mjs.

An approach for sending defferential code will look something like.

```
var legacyExtractor = new ChunkExtractor({ statsFile: legacyStatsFile })
var modernExtractor = new ChunkExtractor({ statsFile: modernStatsFile })
// cache will call the function only once.
let tags = await cache(req.url, () => {
                let legacyJSX = legacyExtractor.collectChunks(<App />);
                let modernJSX = modernExtractor.collectChunks(<App />);
                ReactDOMServer.renderToString(legacyJSX);
                ReactDOMServer.renderToString(modernJSX);
                return {
                    legacyTags: legacyExtractor.getScriptTags(),
                    modernTags: modernExtractor.getScriptTags()
                }
            });
let bodyStream = ReactDOMServer.renderToNodeStream(<App />);
            bodyStream.pipe(res, {end: false});
            bodyStream.on('end', () => {
                res.end(`</div>${
                        tags.legacyTags
                        .replace(/<script/g,"<script nomodule")
                    }${
                        tags.modernTags
                        .replace(/<script/g,'<script type="module"')
                    }
                </body></html>`);
            });
```
It will be much better in next version if lodable-component itself will return script tags with type="modules" and nomodules.

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -9,6 +9,7 @@ import { smartRequire, joinURLPath } from './util'
 
 const EXTENSION_SCRIPT_TYPES = {
   '.js': 'script',
+  '.mjs': 'script',
   '.css': 'style',
 }
 


neoziro:
Hello @ashuanindian thanks for your PR. Yeah I know that ES Modules are coming, I would like to add support to be able to insert in the page (see https://twitter.com/addyosmani/status/1082902291328360448). I will not merge it today but we will work on it.

luisherranz:
Hey guys, we are working on this as well, although we are using a different approach.

We do not use `renderToString` twice, we manually import the `loadable-stats-x.json` files generated by the module and es5 webpack bundles and we use non-public APIs `@loadable/server` to get the info we need and then search for the correct assets of each chunk:

```js
const moduleStats = await import('./build/loadable-stats-module.json');
const es5Stats = await import('./build/loadable-stats-es5.json');

const assets = extractor.getMainAssets("script").map(chunk => chunk.chunk);

const moduleTags = assets.map(
  chunk =>
    `<script async type="module" data-chunk="${chunk}" src="${extractor.publicPath}/${
      moduleStats.assetsByChunkName[chunk]
    }"></script>`
);

const es5Tags = assets.map(
  chunk =>
    `<script async nomodule data-chunk="${chunk}" src="${extractor.publicPath}/${
      es5Stats.assetsByChunkName[chunk]
    }"></script>`
);

const requiredChunksTag = extractor.getRequiredChunksScriptTag({});

const scriptTags = [requiredChunksTag, ...moduleTags, ...es5Tags].join("\n");
```

It'd be amazing if you could provide:
 - A public function to get an array of chunks used: for example `extractor.getScriptStrings()`.
 - Move `extractor.publicPath` to the public API.
 - Move `extractor.getRequiredChunksScriptTag()` to the public API.

I can work on the PR if you are willing to accept this changes.

neoziro:
Hey @luisherranz, thanks for your feedback, I would like to provide a simple way for users to be able to integrate this. I think we could simplify the process for users. The API should not move and it should provide support for module and non-module scripts. For that we need to know which one is a module and which one is not. I have to dig in the webpack and babel API to see what we could do.

luisherranz:
Super :)

kamleshchandnani:
Hey @neoziro! Are you planning to merge this anytime soon? I've similar use case where I'm generating 2 variants of bundles and the one which is es module ends with `.mjs`. Is there any work left out so I can contribute?

neoziro:
Hello @ashuanindian, I will try to work on it before august. Not sure I will have time to do it.

neoziro:
Anyway I close the PR, `.mjs` should not be used client side. So the fix looks not good.

PR #244, opened 2019-02-18
Warn against having comments=false in babel

mike-marcacci:
## Summary

It took me an hour+ to figure out why I wasn't getting chunk names in my stats file... turns out babel was stripping the comments injected by this plugin. It isn't immediately obvious how the plugin works, so I figured I'd save the next person some time and make a note here.

## Test plan

Nothing; it's just a documentation change.


neoziro:
Thanks! It will be helpful!

PR #246, opened 2019-02-19
Adding Named Exports support for SSR

### LINKED ISSUE #245: Failing to import named exports server side
## 🐛 Bug Report

When importing a named export from another file, I receive the following error

```
{
  "status": "error",
  "message": "Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.",
  "stack": [
    "Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.",
    "at invariant (webpack:///../node_modules/react-dom/cjs/react-dom-server.node.development.js?:58:15)",
    "at ReactDOMServerRenderer.render (webpack:///../node_modules/react-dom/cjs/react-dom-server.node.development.js?:3395:7)",
    "at ReactDOMServerRenderer.read (webpack:///../node_modules/react-dom/cjs/react-dom-server.node.development.js?:3131:29)",
    "at renderToString (webpack:///../node_modules/react-dom/cjs/react-dom-server.node.development.js?:3598:27)",
    "at eval (webpack:///./src/server/render.jsx?:40:90)","at Layer.handle [as handle_request] (webpack:///../node_modules/express/lib/router/layer.js?:95:5)",
    "at trim_prefix (webpack:///../node_modules/express/lib/router/index.js?:317:13)","at eval (webpack:///../node_modules/express/lib/router/index.js?:284:7)",
    "at Function.process_params (webpack:///../node_modules/express/lib/router/index.js?:335:12)","at next (webpack:///../node_modules/express/lib/router/index.js?:275:10)"
  ]
}
```

## To Reproduce

Express SSR setup

**App.js**
```jsx
import React from 'react'
import loadable from '@loadable/component'
import './main.css'

const A = loadable(async () => {
  const { A: AComponent } = await import('./letters/A')
  return () => <AComponent />
})

const App = () => (
  <div>
    <A />
  </div>
)

export default App
```

**./letters/A.js**
```jsx
export const A = () => 'A'
```

## Expected behavior

Named export should resolve server side and render correctly

## Link to repl or repo (highly encouraged)

[Example Repo](https://github.com/robdonn/ssr-loadable-component-named-export)

robdonn:
## Summary

Fixes #245 

Problem arose while trying to import named exports server side. Method described in #97 didn't work (I suspect because `renderToString` doesn't support `async/await`) so this seemed like the ideal solution.

## Test plan

**OtherComponent.js**

```js
export const OtherComponent = () => 'Other Component'
```

**App.js**

```js
const OtherComponent = loadable(() => import('./OtherComponent'), 'OtherComponent')

function MyComponent() {
  return (
    <div>
      <OtherComponent />
    </div>
  )
}
```

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -20,8 +20,22 @@ const withChunkExtractor = Component => props => (
 const identity = v => v
 
 function createLoadable({ resolve = identity, render, onLoad }) {
-  function loadable(loadableConstructor, options = {}) {
+  function loadable(loadableConstructor, ...additionalArguments) {
     const ctor = resolveConstructor(loadableConstructor)
+    let componentName
+    let options = {}
+
+    if (additionalArguments.length) {
+      if (additionalArguments.length === 2) {
+        [componentName, options] = additionalArguments
+      } else if (additionalArguments.length === 1) {
+        if (typeof additionalArguments[0] === 'string') {
+          [componentName] = additionalArguments
+        } else if (typeof additionalArguments[0] === 'object') {
+          [options] = additionalArguments
+        }
+      }
+    }
 
     class InnerLoadable extends React.Component {
       constructor(props) {
@@ -97,7 +111,7 @@ function createLoadable({ resolve = identity, render, onLoad }) {
         try {
           const loadedModule = ctor.requireSync(this.props)
           const result = resolve(loadedModule, { Loadable })
-          this.state.result = options.named ? result[options.named] : result
+          this.state.result = componentName ? result[componentName] : result
           this.state.loading = false
         } catch (error) {
           this.state.error = error
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -26,7 +26,7 @@ function createLoadable({ resolve = identity, render, onLoad }) {
 
     if (typeof additionalArguments[0] === "object") {
       componentName = undefined;
-      options = additionalArguments[0];
+      [options] = additionalArguments;
     }
 
     class InnerLoadable extends React.Component {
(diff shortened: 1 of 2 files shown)


neoziro:
Hello @robdonn, thanks for your work. Unfortunately, this is not the good solution. What you describe in #245 should work. To make it work we have to improve the babel function, to make it smarter.

I don't want to introduce a "named" option, I think it is confusing. Your solution could be easily solved by an intermediary component.

robdonn:
Hi @neoziro,

Thank you for the feedback. 

I'd be interested to know why you think this is not a good solution.

```js
const ComponentName = loadable(() => import('../Component'), {
  named: 'ComponentName'
})
```

vs

```js
const ComponentName = loadable(async () => {
  const { ComponentName: ComponentImport } = await import('../Component')
  return (props) => <ComponentImport {...props} />
})
```

Other solutions introduces additional complexity to the code that users need to write. Without my above solution a user would need to write intermediary components, or rewrite all of their components to default exports. This would make Loadable Components very opinionated on _how_ users should write their code. The company that I work for, for example, do not allow default exports. ([Here](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/) is an article which goes over a number of reasons why, just for context)

If there is a way to allow this as a configuration option that looks simpler then I will definitely do a rewrite. Maybe have an optional parameter?

e.g.
```js
const ComponentName = loadable(() => import('../Component'), 'ComponentName', {})
```

I have updated the files to reflect this new solution.

neoziro:
I understand, but we have to choose one way to do thing. Someone could not like `default` and `named` and just say: I exports a plain object in default because I like it better.

I don't see a significant reason to add a specific case for named exports. In fact I follow [the specification established by React](https://reactjs.org/docs/code-splitting.html#named-exports). I suggest you to open a new issue on React, if they change their mind, I will change mine.

robdonn:
@neoziro Thanks for the feedback, I'm reading through the React discussions and it is completely fair for you to keep your library in line with them. 👍

PR #257, opened 2019-03-07
Use inline JSON to enabling CSP without `unsafe-inline`

ambar:
## Summary

See https://mathiasbynens.be/notes/json-dom-csp

## Test plan

Dynamic loading works in `examples/server-side-rendering` and `examples/razle`.


neoziro:
Amazing thanks!

PR #277, opened 2019-03-17
implement attribute function

salzhrani:
## Summary

Allow the attributes arguments to be a function to possibly map the chunks to attributes.

Currently it uses the same argument but checks if it is a function or an object.

Also the attribute function could get `null` as an argument in case the `__LOADABLE_REQUIRED_CHUNKS__` tag

Looking forward to your feedback.

## Test plan

tests included

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -20,9 +20,13 @@ function getAssets(chunks, getAsset) {
   return _.uniqBy(_.flatMap(chunks, chunk => getAsset(chunk)), 'url')
 }
 
-function extraPropsToString(extraProps) {
-  return Object.keys(extraProps).reduce(
-    (acc, key) => `${acc} ${key}="${extraProps[key]}"`,
+function handleExtraProps(asset, extraProps) {
+  return typeof extraProps === 'function' ? extraProps(asset) : extraProps
+}
+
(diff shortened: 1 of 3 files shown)


packages/server/src/ChunkExtractor.js
```diff
@@ -248,11 +253,12 @@ class ChunkExtractor {
 
   getRequiredChunksScriptTag(extraProps) {
     return `<script id="${LOADABLE_REQUIRED_CHUNKS_KEY}" type="application/json"${extraPropsToString(
+      null,
       extraProps,
     )}>${this.getRequiredChunksScriptContent()}</script>`
   }
 
-  getRequiredChunksScriptElement(extraProps) {
+  getRequiredChunksScriptElement(asset, extraProps) {
```
neoziro: I don't understand this change. Why changing this signature, you can just call `handleExtraProps(null, extraProps)`.

salzhrani: you are correct

neoziro:
@salzhrani any news?

salzhrani:
@neoziro I just added some documentation.

neoziro:
@salzhrani thanks, one last thing, could you rebase please?

salzhrani:
rebased

PR #290, opened 2019-03-28
docs: add note that `@loadable/babel-plugin` is not only needed in SSR

Simek:
## Summary

After migrating from `loadable-components` package I have noticed that even I'm not using SSR but `@loadable/babel-plugin` (`loadable-components/babel` earlier) is required in my setup. 

This change extends Installation note by exposing that `babel-plugin` is also needed when you use chunks.


neoziro:
Thanks for precision!

PR #298, opened 2019-04-08
Improve lodash imports to reduce size of serverless bundles

luisherranz:
## Summary

Serverless is becoming popular and bundle size is important. This PR avoids importing the whole lodash code, even in the server.

## Test plan

Tests pass for `ChunkExtractor.js`, the only affected file.

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -1,7 +1,9 @@
 /* eslint-disable react/no-danger */
 import path from 'path'
 import fs from 'fs'
-import _ from 'lodash'
+import uniq from 'lodash/uniq'
+import uniqBy from 'lodash/uniqBy'
+import flatMap from 'lodash/flatMap'
 import React from 'react'
 import { invariant, LOADABLE_REQUIRED_CHUNKS_KEY } from './sharedInternals'
 import ChunkExtractorManager from './ChunkExtractorManager'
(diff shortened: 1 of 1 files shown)


neoziro:
I like that, thanks @luisherranz!

luisherranz:
That was fast. Thanks @neoziro! :)

luisherranz:
@bitttttten it works if you build your bundle using [webpack](https://webpack.js.org/) or [ncc](https://github.com/zeit/ncc) but I agree that maybe adding them as packages may cover more cases.

bitttttten:
Ohh I see. I am not using webpack to bundle my lambdas! To be honest I didn't ever think of doing that. I guess that answers my question, thanks!

If you bundle your lambda with webpack, then wouldn't you loose all the niceties of the console editor? Or does it some how keep all that?

luisherranz:
I'm not familiar with AWS because I use other service. If code don't need to be bundled, then maybe it's not a good idea to make it necessary if someone is using loadable component.

Could you please describe what is the normal workflow in AWS?

ScriptedAlchemy:
use webpack alias to resolve dependencies

theKashey:
use webpack [externals](https://www.npmjs.com/package/webpack-node-externals) to keep `node_modules` in `node_modules`

ScriptedAlchemy:
Can you run require on a lambda? And require node modules on that side?

theKashey:
Of course, you can. Majority of lambdas(they were not created just for SSR task) are written in pure `nodejs`, and haven't ever experienced any bundler.

PR #299, opened 2019-04-08
add s to speed and reduce so copy is gramatically correct

mobot11:
## Summary

The current use of `speed` and `reduce` on the code-splitting page of the documentation is incorrect. To be grammatically correct, I've added an `s`. 

## Test plan

No tests needed as it is just a grammar update to the documentation 😃


neoziro:
Thanks!

Issue #301, opened 2019-04-08
Absolute node_modules paths are not getting resolved during SSR


fatemab:
## 💬 Questions and Help

I am not sure if this issue is similar to #293.
I am importing a component from node_modules. 
`const ComponentName = loadable(() =>
        import("@nodeModuleName/[path-to-component]")
);`

This import runs on client side. (i.e., when I run using { ssr: false }, the component gets load). But it does not load on server side. Throws error 
`Cannot find module '/Users/[my-project]/@nodeModuleName/[path-to-component]'`

the import gets resolved to
```
var ComponentName = (0, _component.default)({
  chunkName: function chunkName() {
    return "ComponentName";
  },
  isReady: function isReady(props) {
    if (typeof __webpack_modules__ !== 'undefined') {
      return !!__webpack_modules__[this.resolve(props)];
    }

    return false;
  },
  requireAsync: function requireAsync() {
    return import(
    "@nodeModuleName/[path-to-component]");
  },
  requireSync: function requireSync(props) {
    var id = this.resolve(props);

    if (typeof __webpack_require__ !== 'undefined') {
      return __webpack_require__(id);
    }

    return eval('module.require')(id);
  },
  resolve: function resolve() {
    if (require.resolveWeak) {
      return require.resolveWeak("@nodeModuleName/[path-to-component]");
    }

    return require('path').resolve(__dirname, "@nodeModuleName/[path-to-component]");
  }
});
```
Issue #293 mentioned that he is not using webpack but I do use webpack, then also getting error.
Any help is appreciated

theKashey:
so - does `/Users/[my-project]/@nodeModuleName/[path-to-component]` exists, or path is a bit different?

fatemab:
The path should be `/Users/@nodeModuleName/[path-to-component]` i.e., pointing the absolute path of node_modules. but instead the resolve function considers this node_modules absolute path, a relative path and appends it to `path/_dirname`. How should I inform the loadable components that this is an absolute path to node_modules?

neoziro:
I think the problem is here:

```js
return require('path').resolve(__dirname, "@nodeModuleName/[path-to-component]");
```

We should replace it by:

```js
return eval('require.resolve')("@nodeModuleName/[path-to-component]");
```

What do you think @theKashey?

theKashey:
Yes, `require.resolve` is a more proper way especially for this case, where a module is defined above the project.
[...]

PR #303, opened 2019-04-10
fix(babel-plugin): Use require.resolve instead of relative path resolution

fatemab:
## Summary

Node modules were incorrectly getting resolved to relative path during server side rendering.
Reference - [Issue #301](https://github.com/smooth-code/loadable-components/issues/301)

## Test plan
If your import statement is
`const ComponentName = loadable(() => import("some_npm_module") );`
in a file located at -
`~/my-project/src/client/components/my-component.js`

Node was resolving the import path to this relative path :-
`~/my-project/src/client/components/some_npm_module.js`

But this PR will fix this behavior to resolve to its correct path :-
`~/my-project/node_modules/some_npm_module.js`

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -35,7 +35,7 @@ exports[`plugin Magic comment should remove only needed comments 1`] = `
       return require.resolveWeak(\\"moment\\");
     }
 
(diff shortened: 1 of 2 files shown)


fatemab:
> You have to use `eval('require.resolve')` to avoid webpack warnings please.

Done.

neoziro:
Thanks!

Issue #285, opened 2019-03-23
Make loadable.preload() return a promise


luishdz1010:
## 🚀 Feature Proposal
Make loadable.preload() return a promise

## Motivation

I'm implementing a "prefetcher" that will, depending on some conditions, request a bunch of known loadable components from an "index" object whose keys are strings and values loadable-components. At some point, I call preload on a bunch of these and I need a way to await on the loading promise of all components, seems like the only change necessary is to add a `return` statement at the end of the `preload()` function.

## Pitch

To be honest, I kind of expected this to work, I was surprised `preload()` returns undefined.

theKashey:
There is a long discussion, and the solution you are looking for - https://github.com/smooth-code/loadable-components/pull/226

neoziro:
Hello @luishdz1010, `.preload()` is a "preload" action so it can't return a promise. If we expose this API, it will be under `.load()`. But I think it is not the good approach to take, with "Suspense" you will not be able to wait for component to be loaded this way. I suggest you to create your own load method if you want to do it. I prefer to wait for Suspense before adding this kind of API.

luishdz1010:
My bad for not looking hard enough 🤦‍♂️, thanks for the quick reply.

PR #305, opened 2019-04-11
Allow waiting on a preload

crisvergara:
## Summary
Allow waiting on preload.

My main use case is to prevent displaying a fallback when routing between pages. Example:

```
const wait = ms => new Promise((resolve) => setTimeout(resolve, ms));
await Promise.race([MyLoadableComponent.preload(), wait(100)])
showPage();
```

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -192,7 +192,7 @@ function createLoadable({ resolve = identity, render, onLoad }) {
       if (typeof window === 'undefined') {
         throw new Error('`preload` cannot be called server-side')
       }
-      ctor.requireAsync(props)
+      return ctor.requireAsync(props).then(() => {})
     }
 
     return Loadable


theKashey:
🤪 #285, #226 😭

neoziro:
I should reconsider it, if it happens it will be a `load` method, but I am afraid that it is not the philosophy recommended by Suspense.

crisvergara:
I understand you're hemming close to React, but I guess this is a case where I find myself strongly disagreeing with their philosophy. Flashing a loading state is bad, but the `p-min-delay` solution of forcing a loading state is not a real fix either. Both create a navigation experience that looks unprofessional.

I'm going to use the work-around specified here: https://github.com/smooth-code/loadable-components/pull/226#issuecomment-467387019

neoziro:
Yeah it is a good workaround. The best workaround would be to use the fallback. If the fallback is displayed, then you are waiting. When it is hidden, then you are done.

That being said, you could use a `useEffect` in the fallback that notify the « wait » and the « done ».

Using this method, you could wait for anything, data, nested loaded components...

But yes you have to render the component to know if you are waiting or not. In a project, I did it with a hidden router. I should probably release something about it.

PR #313, opened 2019-04-18
Return promise from component.preload()

parties:
**Preface:** I'm not sure if this is the "correct" solution to my problem (maybe I missed something in the documentation) but this was the only way I could find to hook into the process that actually loads the component.

## Summary

The problem I've been experiencing is that I needed to be able to call `Component.preload()` to load a component, but _also **wait** for `preload()` to complete_ **before** rendering the route.

(We're stuck using `react-router` `3.x` for the time being...)

This PR solves our problem by simply returning `ctor. requireAsync(props)` from the `preload()` function, as `requireAsync` already returns a promise that resolves once the component is loaded.

## Test plan

Where are the tests for `loadable/component`? Am I missing something?

### DIFF
--- a/packages/component/package.json
+++ b/packages/component/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@loadable/component",
   "description": "React code splitting made easy.",
-  "version": "5.7.0",
+  "version": "5.8.0-alpha.0",
   "main": "dist/loadable.cjs.js",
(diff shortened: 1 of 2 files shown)


parties:
Closing because I'm bad at searching:
* https://github.com/smooth-code/loadable-components/issues/285
* https://github.com/smooth-code/loadable-components/pull/226

PR #317, opened 2019-04-23
Multiple React applications on same page support

### LINKED ISSUE #311: Support for multiple loadable components on same page
Hello everyone,
    First of all I would like to thank you for such a great library.

We have a requirement where we would like to build a component service. The user can make a request for a particular component and the server returns the html and assets required to render it. 

In the server we run `extractor.collectChunks` and return the things needed for a particular component. I wanted to know if we can have multiple chunks collected in this way from the same server be used on the client. Some problems that I am anticipating:
-  `extractor.getScriptTags()` returns a script tag with `id="__LOADABLE_REQUIRED_CHUNKS__`. This looks like it won't scale for multiple components on the same page
- In `@loadable/component` we have [this](https://github.com/smooth-code/loadable-components/blob/005913be28c6ea197984de4ef5a6d3cd6b96efd6/packages/component/src/loadableReady.js#L34) code which checks if that script tag is loaded before hydrating the component

I wanted to know if my assumptions are right that multiple components on same page are not supported by loadable. If yes then I wanted to know if you would be interesting in including this feature in loadable. I would be willing to help implement this feature in that case (with some guidance of course).

debugpai:
Fixes #311 

## Summary

Loadable components currently doesn't support having multiple react apps on the same page. This is because it uses the hardcoded id of `__LOADABLE_REQUIRED_CHUNKS__`  in a script tag to check if a chunk has loaded. This PR aims to add the feature to add a namespace to each app to make this possible. 

## Test plan

Added test cases on the server side to make sure that the namespace was being added to the `script` id. 
On the client side, verified manually if loading 2 different react apps triggers 2 different `loadable` events

### DIFF
--- a/packages/component/src/loadableReady.js
+++ b/packages/component/src/loadableReady.js
@@ -1,11 +1,14 @@
 /* eslint-disable no-underscore-dangle, camelcase */
 /* eslint-env browser */
 import { warn } from './util'
-import { LOADABLE_REQUIRED_CHUNKS_KEY } from './sharedInternals'
+import { getRequiredChunkKey } from './sharedInternals'
 
 const BROWSER = typeof window !== 'undefined'
 
-export default function loadableReady(done = () => {}, appName = '') {
+export default function loadableReady(
+  done = () => {},
+  { namespace = '' } = {},
+) {
   if (!BROWSER) {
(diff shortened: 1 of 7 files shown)


neoziro:
Thanks for your work!

PR #320, opened 2019-04-23
Making preload built-in chainable

lorenzomigliorero:
Return a `promise` instead of undefined from preload built-in for easy `chaining`.


neoziro:
It is the third PR about that, `preload` does not have to return a promise. I plan to expose a `load` method because it is a hot feature. A lot of people are asking for it.

lorenzomigliorero:
I don't understand why preload doesn't have to return a promise.
What will be the differences between `preload` and `load`?

neoziro:
"preload" means "hey, please preload and it will eventually be loaded", "load is actually the action of loading". Today "preload" acts as "load", but in future I would like to be able to make it work as `<link rel="preload" />`.

PR #329, opened 2019-05-08
feat: add `load` method that returns a Promise

neoziro:
See #226 for context.

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -195,6 +195,13 @@ function createLoadable({ resolve = identity, render, onLoad }) {
       ctor.requireAsync(props)
     }
 
+    Loadable.load = props => {
+      if (typeof window === 'undefined') {
+        throw new Error('`load` cannot be called server-side')
+      }
+      return ctor.requireAsync(props)
+    }
+
     return Loadable
   }
 
--- a/website/src/pages/docs/api-loadable-component.mdx
+++ b/website/src/pages/docs/api-loadable-component.mdx
@@ -48,7 +48,7 @@ A component created using `loadable` or `lazy`.
 
 ## LoadableComponent.preload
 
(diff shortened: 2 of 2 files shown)


neoziro:
@theKashey could you review it please?

packages/component/src/createLoadable.js
```diff
@@ -195,6 +195,13 @@ function createLoadable({ resolve = identity, render, onLoad }) {
       ctor.requireAsync(props)
     }
 
+    Loadable.load = props => {
+      if (typeof window === 'undefined') {
+        throw new Error('`load` cannot be called server-side')
```
theKashey: It is really has to be an error?

theKashey:
😍, welcome to the club

packages/component/src/createLoadable.js
neoziro: Another issue, for now I will let it like this.

bertho-zero: I was happy to see this PR but looking at the code I can not use it because of this limitation.  I need as much client side as server.

neoziro: I removed it 🤗

bertho-zero: Thank you, that would be perfect with a method like isReady.

neoziro: Yep, but remember, if you need it, you are probably taking the wrong way to be Suspense ready. When Suspense will be released, you will not be able to wait for anything except in render.

bertho-zero: I know that Suspense does not allow it, but it's still useful information. In my case it allows me to display a progress bar when you have to load one or more chunks and not display it when they are already loaded.

theKashey: Ooff, how many chunks you have to load?

PR #330, opened 2019-05-08
feat: support reactive dynamic loadable

### LINKED ISSUE #284: Fully dynamic loadable is not reactive
## 🐛 Bug Report
Seems like only the very first prop sent to the loadable component in a [dynamic]( https://www.smooth-code.com/open-source/loadable-components/docs/dynamic-import/) context is respected, further updates to the props have no effect. Same applies to loadable.lib

## To Reproduce
https://codesandbox.io/s/3wn6vo7ym

## Expected behavior
Changing props will work reactively. An "easy" fix is to set a `key` prop to the loadable component, this works in the case of using a loadable component (although it seems like a hack) but for loadable.lib, this will trigger an unmount and remount of all children of the component, which is sometimes not desired as the rendered tree could be similar, you want to perform animations, etc.

```
## System:
 - OS: macOS 10.14.3
 - CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
 - Memory: 99.34 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 10.15.2 - /usr/local/opt/node@10/bin/node
 - Yarn: 1.15.2 - /usr/local/bin/yarn
 - npm: 6.4.1 - /usr/local/opt/node@10/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
## npmPackages:
 - @loadable/babel-plugin: ^5.7.2 => 5.7.2
```

neoziro:
Closes #284

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,6 +8,7 @@
   "rules": {
     "no-plusplus": "off",
     "no-param-reassign": "off",
+    "no-nested-ternary": "off",
     "react/jsx-filename-extension": ["error", { "extensions": [".js"] }],
     "react/jsx-wrap-multilines": "off",
     "react/no-unused-state": "off",
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,13 +1,14 @@
-let targets = {}
-
-if (process.env.BUILD_TARGET === 'node') {
-  targets = { node: '8' }
+function getTargets() {
+  if (process.env.BUILD_TARGET === 'node') {
+    return { node: '8' }
+  }
+  return undefined
 }
 
 module.exports = {
   presets: [
     ['@babel/preset-react', { useBuiltIns: true }],
-    ['@babel/preset-env', { loose: true, targets }],
+    ['@babel/preset-env', { loose: true, targets: getTargets() }],
   ],
   plugins: ['@babel/plugin-proposal-class-properties'],
 }
--- a/examples/client-side/package.json
+++ b/examples/client-side/package.json
@@ -6,13 +6,13 @@
   "devDependencies": {
     "babel-loader": "^8.0.4",
     "html-webpack-plugin": "^3.2.0",
-    "webpack": "^4.23.1",
-    "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "^3.1.10"
+    "webpack": "^4.30.0",
+    "webpack-cli": "^3.3.2",
+    "webpack-dev-server": "^3.3.1"
   },
   "dependencies": {
-    "moment": "^2.22.2",
-    "react": "^16.6.0",
-    "react-dom": "^16.6.0"
+    "moment": "^2.24.0",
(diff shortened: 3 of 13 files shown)


neoziro:
@theKashey could you review it please?

theKashey:
I would like to make it a bit more complex:
- an option to display old Component while a new one is loading. It could be an option to a LoadingComponent. The problem - right now you are loosing `cacheKey` and unable to refer to a prev data
- if `cacheKey` got changed during loading - it would not cancel `loadAsync` and it would update component. You have to add more checks (more about store ref to a promise created, and compare it with the current one).
- _theoretically_ you shall not display `old` component with a `new` props, if a `new` component should be displayed instead. What about using `getDerivedStateFromProps` to derive `state.props` from `props`, and __freeze__ them, if `state.usedCacheKey`!==`state.cacheKey`. 

This would be needed only for `.1`, as long as it's much easier to add some cache-key logic in a render (still having old `cache-key` in a `state`). But, in the same time, this logic would ruin all the structure, unless implemented in hooks.

Oh, it really could become a bit complicated.

neoziro:
> an option to display old Component while a new one is loading. It could be an option to a LoadingComponent. The problem - right now you are loosing cacheKey and unable to refer to a prev data

It is not the responsibility of this component, it could be done in userland I think.

> if cacheKey got changed during loading - it would not cancel loadAsync and it would update component. You have to add more checks (more about store ref to a promise created, and compare it with the current one).

Yeah you are right!

> theoretically you shall not display old component with a new props, if a new component should be displayed instead. What about using getDerivedStateFromProps to derive state.props from props, and freeze them, if state.usedCacheKey!==state.cacheKey.

Yes you are right, I think `getDerivedStateFromProps` is a good workaround.

luishdz1010:
Glad this got in!

I'm trying to implement keeping the old component mounted while loading in userland and it works fine, however one issue I cannot seem to move past is that there is always a `loading` phase, even if both components have already loaded.

I'd expect a seamless switch between them, like:

```
CompB -> CompB
```

But currently it goes: 

```
CompA -> fallback (loading === true) -> CompB
```

Even if both CompA and CompB were previously loaded. If using loadable.lib, then the children-as-a-function prop gets "skipped" for a brief moment until the next render (which happens immediately). This causes problems with animations because it unmounts/mounts the whole tree under loadable.lib.

Just wanted to point this out for future readers seeking a solution to this, I understand fixing this may be a bit more complex than acceptable.

neoziro:
Hello @luishdz1010, for now, yes it will switch to `loading` state for a very small time. But even if it was not the case, the component will still be mounted / unmounted and it will not solve your animation problem.

To display the previous component during loading, the strategy will be the same as a [Transition Group](https://reactcommunity.org/react-transition-group/). You have to check when the component will change and keep displaying the previous one until the new one is ready (fallback has disappeared).

PR #370, opened 2019-06-29
fix: add Object.is polyfill to improve ie compatibility

NaccOll:


theKashey:
No. No library ever should polyfill anything, ie apply a side effect on a global object.
Ponyfill - you are welcome.

NaccOll:
close by #370

PR #375, opened 2019-07-03
Generates chunk names consistently

### LINKED ISSUE #373: Multiple chunks for same component when paths different
## 🐛 Bug Report

When we call `loadable` and import the same component from different places using different paths then 2 different chunks are created. 

## To Reproduce

Example consider the folder structure like this
- index.js
- src/
---- LazyComponent.js
---- Simple.js

In `index.js`
```js
  const LoadableLazy = loadable(() => import('./src/LazyComponent'));
  return (
     </div>
       <LoadableLazy />
        <Simple />
    </div>
```
In `Simple.js`
```js
  const LoadableLazy = loadable(() => import('./LazyComponent'));
  return <LoadableLazy />;
```

Here 2 different chunks namely `src-LazyComponent` and `LazyComponent`

## Expected behavior
Only 1 chunk should be created

I would like to pick this issue up. I think changing [this](https://github.com/smooth-code/loadable-components/blob/master/packages/babel-plugin/src/properties/chunkName.js#L76) function to generate chunk names based on either content hash or relative path with respect to project root would solve this issue.

Would like some guidance to fix this.

debugpai:
## Summary
Fixes #373. Uses the path relative to the project root when generating chunkNames instead of using the import path directly.

## Test plan
Added a new test in `babel-plugin/src/index.test.js`.
I updated the snapshots and checked the new webpackChunkNames

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -491,7 +491,7 @@ exports[`plugin simple import should transform path into "chunk-friendly" name 1
 exports[`plugin simple import should transform to same chunk name when import paths are different 1`] = `
 "loadable({
   chunkName() {
-    return \\"Users-dthonse-workspace-loadable-components-baz-foo-bar\\";
+    return \\"home-travis-build-smooth-code-loadable-components-baz-foo-bar\\";
   },
 
   isReady(props) {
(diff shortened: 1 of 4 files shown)


bitttttten:
Could this also resolve https://github.com/smooth-code/loadable-components/issues/382?

debugpai:
That seems to be a different issue. This PR only deals with static imports.

bitttttten:
Alright, thanks!

neoziro:
Hello @debugpai, I am sorry your method will not work with webpack aliases. To solve your issue you have to use `webpackChunkName` we can't solve it in Loadable Components.

PR #416, opened 2019-08-22
support ie11

### LINKED ISSUE #397: IE11: Object doesn't support property or method 'includes'
## 🐛 Bug Report

Hi!
According to docs browser support is:
> v5.x (React v16.3+): IE11, IE 9+ (with Map + Set polyfills), all evergreen browsers

But my project fails on this line https://github.com/smooth-code/loadable-components/blob/d7c3f58059a798dc1d38f95da79f374e998977ad/packages/component/src/loadableReady.js#L44

`includes` doesn't supported in IE up to ie11

Am i doing something wrong? My project uses `razzle` boilerplate with `loadable-components` example: https://github.com/smooth-code/loadable-components/tree/master/examples/razzle

Haroenv:
I didn't check this, but it seems trivial

fixes #397

### DIFF
--- a/packages/component/src/loadableReady.js
+++ b/packages/component/src/loadableReady.js
@@ -41,7 +41,7 @@ export default function loadableReady(
     function checkReadyState() {
       if (
         requiredChunks.every(chunk =>
-          loadedChunks.some(([chunks]) => chunks.includes(chunk)),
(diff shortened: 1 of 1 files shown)


neoziro:
Thanks @Haroenv!

PR #417, opened 2019-08-25
Add required dependencies to run the razzle example

Caerbannog:
## Summary

The Razzle example is missing @loadable/component and @loadable/server in package.json to run.

## Test plan

With this commit, running the following commands should not fail anymore:
```
cd examples/razzle
npm install
npm run build # or start
```


neoziro:
Thanks!

PR #424, opened 2019-09-03
render errorComponent if ctor promise rejected

XaveScor:
Hello. I have a case that I cannot solve with loadable-components.
Case:
If I cannot download some chunk I need to show the popup to user. But popup's background should contain already downloaded part of site. If I try to catch error with React Error Boundaries my js script contains an infinite loop because I rerender React tree with invalid chunk.

My solution allows to fix this edge case. I can put some component with React.Portal via options, for example.

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -194,6 +194,9 @@ function createLoadable({ resolve = identity, render, onLoad }) {
         }
 
         if (error) {
+          if (options.error !== undefined) {
+            return options.error
+          }
           throw error
         }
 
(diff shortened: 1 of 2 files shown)


theKashey:
`ErrorBoundary` should be enough to implement any logic you need. Just don't try to render loadable back until user will ask for it (by clicking a retry button)

XaveScor:
How to implement my logic? How to save healthy React Tree? Healthy = all tree except rejected chunks.

neoziro:
To save your tree, put an error boundary around your Loadable Component and only around it. It is an opinionated choice to use only Error Boundaries.

PR #425, opened 2019-09-04
Babel plugin: Fixed bug when using + concatenation instead of a template literal

mbrowne:
## Summary

Currently, the Babel plugin successfully transpiles template literals, but doesn't work when building a string using `+`. i.e. this works:

```js
import(`my-lib/${componentId}`)
```

But this does not:
```js
import('my-lib/' + componentId)
```

The latter fails with an error:
> TypeError: Property value expected type of string but got null

This PR fixes this issue by checking to see if the import expression is a BinaryExpression. I'm not sure if there are any other types of expressions that should be checked for, but this solves the issue for BinaryExpressions at least.

## Test plan

I added a new test to verify that it works and also tested it in a real app.

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -527,6 +527,45 @@ exports[`plugin simple import should work with * in name 1`] = `
 });"
 `;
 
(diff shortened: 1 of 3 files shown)


theKashey:
👍

neoziro:
Thanks!

PR #435, opened 2019-09-22
Improved docs for ssr and code splitting

aganglada:
Added documentation for 

- Code splitting (named imports, client and server side).
- Server Side Rendering entrypoints.

## Summary

In order to share the knowledge from a problem I've run into recently, I would like to write this points down in the documentation to help people don't make the same mistakes as me.

### DIFF
--- a/website/src/pages/docs/code-splitting.mdx
+++ b/website/src/pages/docs/code-splitting.mdx
@@ -43,6 +43,25 @@ If you’re setting up Webpack yourself, you’ll probably want to read [Webpack
 
 When using Babel, you’ll need to make sure that Babel can parse the dynamic import syntax but is not transforming it. For that you will need [@babel/plugin-syntax-dynamic-import](https://www.npmjs.com/package/@babel/plugin-syntax-dynamic-import).
 
+## Named `import()`
+
+Weckpack's default behaviour, is to name them as `0.js` where 0 is an incremental number depending on how many dynamic
+chunks you are importing in your code.
+
+This will give a very poorly view of which file is loading what code.
+
+To fix that, webpack introduced magic comments, with which a chunck can be named as follows (ie: `math.js`). 
+
+```js
+import(/* webpackChunkName: "math" */ './math').then(math => {
+  console.log(math.add(16, 26))
+})
+```
+
+>NOTE: 
+>When using [Server Side Rendering](/docs/server-side-rendering/), make sure comment and file path are in exacly the same order as above.
+
+
 ## Code Splitting + React
 
 React supports code splitting out of the box with [`React.lazy`](https://reactjs.org/docs/code-splitting.html#reactlazy). However it has [some limitations](/docs/loadable-vs-react-lazy), this is why `@loadable/component` exists.
--- a/website/src/pages/docs/server-side-rendering.mdx
+++ b/website/src/pages/docs/server-side-rendering.mdx
@@ -116,6 +116,18 @@ The `extractor.getScriptTags()` returns a string of multiple `<script>` tags mar
 
 Alternatively the `ChunkExtractor` also has a `getScriptElements()` method that returns an array of React elements.
 
+### Entrypoints
+
+By default, `ChunkExtractor` will collect chunks from the `main` entrypoint (which is the default behaviour of `webpack`), if you would
+like to change this, you can pass an `entrypoints` option like so.
+
+```js
+const extractor = new ChunkExtractor({ 
+  statsFile, 
+  entrypoints: ['client'] // name of the proper webpack endpoint (default: main)
+})
+```
+
 ## Streaming rendering
 
 Loadable is compatible with streaming rendering, if you use it you have to include script when the stream is complete.


website/src/pages/docs/code-splitting.mdx
```diff
@@ -43,6 +43,25 @@ If you’re setting up Webpack yourself, you’ll probably want to read [Webpack
...
+To fix that, webpack introduced magic comments, with which a chunck can be named as follows (ie: `math.js`). 
+
+```js
+import(/* webpackChunkName: "math" */ './math').then(math => {
+  console.log(math.add(16, 26))
+})
+```
+
+>NOTE: 
+>When using [Server Side Rendering](/docs/server-side-rendering/), make sure comment and file path are in exacly the same order as above.
```
neoziro: are exactly in

website/src/pages/docs/code-splitting.mdx
```diff
@@ -43,6 +43,25 @@ If you’re setting up Webpack yourself, you’ll probably want to read [Webpack
 
 When using Babel, you’ll need to make sure that Babel can parse the dynamic import syntax but is not transforming it. For that you will need [@babel/plugin-syntax-dynamic-import](https://www.npmjs.com/package/@babel/plugin-syntax-dynamic-import).
 
+## Named `import()`
+
+Weckpack's default behaviour, is to name them as `0.js` where 0 is an incremental number depending on how many dynamic
```
neoziro: `x.js` where `x`

website/src/pages/docs/server-side-rendering.mdx
```diff
@@ -116,6 +116,18 @@ The `extractor.getScriptTags()` returns a string of multiple `<script>` tags mar
 
 Alternatively the `ChunkExtractor` also has a `getScriptElements()` method that returns an array of React elements.
 
+### Entrypoints
+
+By default, `ChunkExtractor` will collect chunks from the `main` entrypoint (which is the default behaviour of `webpack`), if you would
```
neoziro: This is not exact, we use `entrypoints` as `assets, we are not collecting chunks from them.

Please remove this section.

theKashey: Well, what [getMainAssets](https://github.com/smooth-code/loadable-components/blob/master/packages/server/src/ChunkExtractor.js#L331) is doing then? Main entry point might depend on some extra chunks, and that have to be counted separately from "loadable"-ed chunks.
As long as these "entry" chunks and chunks used by `NamedAssets` could overlap - this sentence is 100% true.

website/src/pages/docs/code-splitting.mdx
aganglada: Fixed, thanks!

website/src/pages/docs/code-splitting.mdx
aganglada: Fixed, thanks!

website/src/pages/docs/server-side-rendering.mdx
aganglada: Okay, let me dig into the problem once more, because with my webpack configuration, loadable didn't work till I specified `entrypoints`. I will get back with feedback, in the meantime, we can discuss.

Thanks for all your feedback.

aganglada: My webpack config entries for client/server are specified this way:

Client
```
entry: {
   client: './src/client.tsx',
},
```

Server
```
entry: {
   server: './src/server.tsx',
},
```

If they were this way:

Client
```
entry: './src/client.tsx'
```

Server
```
entry: './src/server.tsx'
```

It will work because webpack will call those entries `main.js`, but otherwise it errors saying:

```
Invariant Violation: loadable: cannot find main in stats
```

I think this case should be specified in the docs, as it took me several days to find out the problem was coming from `loadable-stats.json` (for some reason I thought `loadable` was getting entries from `webpack` dynamically).

Let me know if you agree on this. 
I will also rephrase this section of the docs being more specific on this case.

theKashey: Many people spend quite a lot of time here. I've asked :)

aganglada: @neoziro @theKashey I've rephrased `Entrypoints` in what I think is a better explanation of the problem.

Let me know what you think and if I need to change something.

Thanks!

website/src/pages/docs/server-side-rendering.mdx
```diff
@@ -116,6 +116,36 @@ The `extractor.getScriptTags()` returns a string of multiple `<script>` tags mar
 
 Alternatively the `ChunkExtractor` also has a `getScriptElements()` method that returns an array of React elements.
 
+### Entrypoints
```
neoziro: This is very advanced, can you please put it at the very end of the page? Thanks!

aganglada: Sure, no problem.

aganglada: I moved the section to the end of the page and modified title to give more context.
Thanks for the feedback.

Issue #344, opened 2019-05-20
Support SRI on webpack-plugin


yayoc:
## 🚀 Feature Proposal

Putting an integrity string for [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) into `loadable-stats.json` when building bundles.

## Motivation

To embed `integrity` attribute for each script tag when running SSR application with loadable components, I need to read the file and compute the integrity value on runtime. 
Instead of that, I would like to compute the integrity value at the building phase to reduce overhead on runtime.

## Example

Putting integrity value into `loadable-stats.json` 

```json
{
      "name": "xxx-e63d1f6f130186a006da.js",
      "size": 782370,
      "integrity": "sha384-oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC", 
      "chunks": [],
      "chunkNames": [],
      "emitted": false
},
```

Then it could be great if we can access the value on [attrFn for getScriptTags](https://www.smooth-code.com/open-source/loadable-components/docs/api-loadable-server/#chunkextractorgetscripttags)

neoziro:
Hello @yayoc, actually `loadable-stats.json` is just webpack stats, do you know if webpack plans to include it? I do not want to generate it, it is webpack business. If it is available, we would get it for free in SVGR for every script!

yayoc:
@neoziro 
I am not sure if webpack has a plan. Currently, I only know [this plugin](https://www.npmjs.com/package/webpack-subresource-integrity) to generate integrity. 
Do you think if `loadable webpack plugin` can integrate with `webpack-subresource-integrity`?

neoziro:
@yayoc I think it is a great addition! We could add an option and load the plugin on demand if the user needs it. Like an optionalPeerDependency (even if it does not exist) what do you think? We could also make it as default but I don't know if integrity is really useful for your own resources (not pointed on a CDN)?

yayoc:
I agree with loading the plugin as optional for the first step since integrity seems not so high demand. 
@neoziro

peeter-tomberg:
Has there been any progress regarding this?

neoziro:
Hello @peeter-tomberg, no progress. You are welcome to work on it!

PR #436, opened 2019-09-22
Accept additional attributes for assets to support SRI

ashudson23:
## Summary
Potential fix for feature request in issue #344 which will add SRI integrity to be html tags before sending to client, if the value exists on the asset.

### Webpack
```js
const SriPlugin = require('webpack-subresource-integrity');
const LoadablePlugin = require('@loadable/webpack-plugin');

module.exports = ({
  ...
  plugins: [
    ...
    new SriPlugin({
      hashFuncNames: ['sha256', 'sha384'],
    }),
    new LoadablePlugin(),
  ],
});
```

### In the server renderer script
Also worth adding the `crossorigin` key (this functionality already existed and isn't being introduced by this PR)
```js
chunkExtractor.getScriptTags({ crossorigin: 'anonymous' })
```

### Start application
Confirm that during SSR run the tags have `integrity` key with correct value
```html
<script
    async
    data-chunk="HelloWorld"
    src="/1.bundle.js"
    crossorigin="anonymous"
    integrity="sha256-mYmvbcbExknkCdg+X/j/8zsdpRcm29eJ3lpF4wfnl+4= sha384-gECBjN17ng81O/ueB9Dz8Vor95n36djp9M/oWAXg7+4qgCPArWkzSqlEchJlQK3j"
></script>
```

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -332,24 +332,6 @@ class ChunkExtractor {
     return smartRequire(mainAsset.path)
   }
 
-  getSriFromFileName(filename) {
-    if (!filename) {
-      return null
-    }
-
-    const asset = this.stats.assets.find(x => x.name === filename)
-    if (!asset) {
-      return null
-    }
-
-    const { integrity } = asset
-    if (!integrity) {
-      return null
-    }
-
-    return { integrity }
-  }
-
   // Main assets
 
   getMainAssets(scriptType) {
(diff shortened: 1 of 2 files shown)


packages/webpack-plugin/src/index.js
```diff
@@ -26,6 +33,24 @@ class LoadablePlugin {
       errorDetails: false,
       timings: false,
     })
+
+    // include extra asset keys if present
+    stats.assets = stats.assets.map((asset) => {
```
neoziro: I think we will be fine without that isn't it?

ashudson23: Actually, you're absolute right!

packages/server/src/ChunkExtractor.js
```diff
@@ -340,7 +365,10 @@ class ChunkExtractor {
     const requiredScriptTag = this.getRequiredChunksScriptTag(extraProps)
     const mainAssets = this.getMainAssets('script')
     const assetsScriptTags = mainAssets.map(asset =>
-      assetToScriptTag(asset, extraProps),
+      assetToScriptTag({
+        ...asset,
+        ...this.getSriFromFileName(asset.filename),
```
neoziro: All of these is useless, the integrity should already be in asset. The only needed part is : getSriHtmlAttributes and all usage of it.

PR #440, opened 2019-09-26
421 update ssr example

MikeAlexMartinez:
Am in the process of adding code splitting to our Server rendered app at work and wanted to try this out so thought I'd fix the dependencies in the SSR example so they work

motivation is there was a ticket for it already... https://github.com/smooth-code/loadable-components/issues/421

I'm happy to do the others if you would like. Additionally, our app at work uses redux and post css so might make another example trying to get a basic version of this working if you are down for that.

See readme for instructions to get running


neoziro:
@MikeAlexMartinez thanks! Tests are not passing, I don't know why. Anyway I could probably merge it.

MikeAlexMartinez:
Yeah, I though that was strange that the tests failed. I wonder if me adding another markdown file had an affect?

I will be having a look at this over the weekend so maybe I can help figure out the issue 👍

PR #442, opened 2019-10-01
[doc] use webpackPreload hint in example

Swaagie:
## Summary

Small update to the example in the documentation listing `webpackPrefetch` when `webpackPreload` is intended as far as I can tell.

## Test plan

N/A

### DIFF
--- a/website/src/pages/docs/prefetching.mdx
+++ b/website/src/pages/docs/prefetching.mdx
@@ -42,10 +42,6 @@ function App() {
     </div>
   )
 }
-
-const OtherComponent = loadable(() =>
-  import(/* webpackPreload: true */ './OtherComponent'),
-)
 ```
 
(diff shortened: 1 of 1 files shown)


theKashey:
It's might be not a good idea to _preload_ something you want to display `onMouseOver` - prefetch sounds more correct.

website/src/pages/docs/prefetching.mdx
```diff
@@ -44,7 +44,7 @@ function App() {
 }
 
 const OtherComponent = loadable(() =>
```
neoziro: You can just remove this, it is not part of this example 😅

Swaagie: Hah good call, I've removed the `OtherComponent` from the example. I now see the changes are in the `Infos.preload()`, makes them stand out more.

Issue #400, opened 2019-08-05
isReady returns true even though bundles are not all loaded


Ephys:
## 🐛 Bug Report

My app will randomly crash depending on the load order of my bundles because `loadable()`'s `isReady` doesn't check that all bundles that are needed are loaded (only the main one is checked).

## To Reproduce

Ok so this is pretty difficult to reproduce because it depends on the load order of different files

I have the following piece of code:

```javascript
const LoginComp = loadable(() => import('../Login'));
```

which is being transpiled by `@loadable/babel-plugin` & `@loadable/webpack-plugin` into this:

```javascript
const LoginComp = Object(_loadable_component__WEBPACK_IMPORTED_MODULE_1__[/* default */ "a"])({
  chunkName: function chunkName() {
    return "Login";
  },
  isReady: function isReady(props) {
    if (true) {
      return !!__webpack_require__.m[this.resolve(props)];
    }

    return false;
  },
  requireAsync: function requireAsync() {
    return Promise.all(/* import() | Login */[__webpack_require__.e(0), __webpack_require__.e(2), __webpack_require__.e(1), __webpack_require__.e(3), __webpack_require__.e(63)]).then(__webpack_require__.bind(null, 250));
  },
  requireSync: function requireSync(props) {
    var id = this.resolve(props);

    if (true) {
      return __webpack_require__(id);
    }

    return eval('module.require')(id);
  },
  resolve: function resolve() {
    if (true) {
      return /*require.resolve*/(250);
    }

    return eval('require.resolve')("../Login");
  }
})
```

In this piece of code, you can see that `requireAsync` first loads the *bundles 0, 2, 1, 3 and 63* before loading *module 250* (module 250 is contained by bundle 63 and depends on code present in 0, 2, 1, 3).

Now, I have a separate piece of code that will eventually cause the whole react tree to re-render as soon as another file is loaded (a workaround for react-router which still uses the old context api - I re-render the react tree once the proper locale file is loaded).

Because the whole react tree re-renders, the component generated by `loadable()` is re-built from scratch.

When that component constructs, it calls `isReady` to check if it should use `requireSync` or `requireAsync`. But `isReady` only checks that bundle 63 is loaded because it only checks for module 250. So if 0, 2, 1, or 3 are not loaded yet, it will load module 250 which will crash with `cannot read property .call of undefined` because module 250 will attempt to load a module from one of the other bundles

## Expected behavior

`isReady` is smart enough to detect whether all modules are loaded or not

[...]

PR #445, opened 2019-10-08
Issue 400

theKashey:
## Summary
Track the loading state for `isReady`, should solve #400 

The logic so:
- there is no `__webpack_modules__[key]` - load async
- there is `__webpack_modules__[key]` - load sync, only if you are still not loading async

In other words - don't load while you load.

## Test plan

Hard to reproduce, you have to load component, which would require more than one chunk to load, and the first one to be loaded before second.


foglerek:
@neoziro Do you think this approach is going to end up getting merged in, or are you planning on waiting on webpack to expose a more reliable sync-load method?

theKashey:
👍

theKashey:
`loading` was not the best name, as long as in case of failed loading it would be still `loading`, while it does nothing.
As a result, it was replaced by `resolved`

iamstarkov:
@theKashey should the tests be fixed?

theKashey:
Sorry, forgot to push updated snapshots

theKashey:
@neoziro - only you can merge and publish

mschipperheyn:
@neoziro @theKashey any chance this can get merged soonish? The main disadvantage for monorepos is that you can't install directly by git commit.

ilyaagarkov:
Hi everyone!
@neoziro @theKashey Any updates?

Issue #225, opened 2019-02-06
loadable-stats.json's are created as duplicates.


hemmxwxsoo:
## 🐛 Bug Report

If I set a `writeToDisk.filename` and it is different from `output.path`, They are created in both places.

## To Reproduce

1. create a files

``` javascript
// ./webpack.config.js
const path = require('path'),
      loadable_plugin = require('@loadable/webpack-plugin');

module.exports = {
    entry: path.join(__dirname, 'src/index.js'),
    output: { filename:'index.js', path:path.join(__dirname, 'dist/static/js') },
    plugins: [ new loadable_plugin({ writeToDisk:{ filename:path.join(__dirname, 'dist') }}) ]
};
```

``` javascript
// ./src/index.js
```

``` javascript
// ./package.json
{
  "version": "1.0.0",
  "description": "",
  "scripts": {
    "build": "webpack",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@loadable/webpack-plugin": "^5.5.0",
    "webpack": "^4.29.2",
    "webpack-cli": "^3.2.3"
  }
}
```

2. install packages
```bash
npm install -D
```

3. build
```bash
npm run build
```

4. You will probably got a two files `./dist/loadable-stats.json` and `./dist/static/js/loadable-stats.json`.

## Expected behavior

I only want to get a `./dist/loadable-stats.json`.

## Link to repl or repo (highly encouraged)

Please provide a minimal repository on GitHub.

Issues without a reproduction link are likely to stall.

## Run `npx envinfo --system --binaries --npmPackages @loadable/component,@loadable/server,@loadable/webpack-plugin,@loadable/babel-plugin --markdown --clipboard`

```bash
## System:
 - OS: Windows 10
 - CPU: (4) x64 Intel(R) Pentium(R) CPU G4600 @ 3.60GHz
 - Memory: 432.83 MB / 7.70 GB
## Binaries:
 - Node: 8.9.1 - C:\Program Files\nodejs\node.EXE
 - Yarn: 1.13.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
 - npm: 6.5.0 - C:\Program Files\nodejs\npm.CMD
```

neoziro:
Hello @hemmxwxsoo, it is actually the expected behaviour. But I would like to create another option to avoid emitting the file.

hemmxwxsoo:
[...]

PR #449, opened 2019-10-20
Document outputAsset in api-loadable-webpack-plugin.mdx

tamj0rd2:
## Summary

I ran into the same issue as #225  and noticed that the outputAsset option is not documented on the website.

## Test plan

N/A


tamj0rd2:
Personally though, I think it's a bit confusing that outputAsset is an option when writeToDisk already exists.

PR #470, opened 2019-12-02
feat: passing chunkName and extractor to Component

luojunbin:
## Summary

I want to get dynamic css chunks before `renderToNodeStream` when SSR.

This PR add 2 parameters(`extractor` and `chunkName`) to the child component wrapped by `loadable`.

## Test plan
```js
// test.js
export default function Test(props) {
    // return 'Test & Assets'
    return `${props.loadableChunkName} & ${props.loadableChunkExtractor.getChunkAssets([props.loadableChunkName])}`;
}

loadable(() => import('./test'))
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 lib/
 /coverage/
 examples/*/yarn.lock
+.idea
--- a/packages/component/.size-snapshot.json
+++ b/packages/component/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/loadable.cjs.js": {
-    "bundled": 12683,
-    "minified": 6125,
-    "gzipped": 2203
+    "bundled": 12941,
+    "minified": 6233,
+    "gzipped": 2237
   },
   "dist/loadable.esm.js": {
-    "bundled": 12300,
-    "minified": 5816,
-    "gzipped": 2135,
+    "bundled": 12558,
+    "minified": 5924,
+    "gzipped": 2166,
     "treeshaked": {
       "rollup": {
         "code": 259,
         "import_statements": 259
       },
       "webpack": {
-        "code": 5068
+        "code": 5176
       }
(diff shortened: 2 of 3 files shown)


theKashey:
Do you really need this?
What if more than one chunk was used to display "this" component?

luojunbin:
I can use a `Set` to ensure that dynamic css chunks only occur once, here is an example

```js
function renderLink(extractor, chunkName) {
  if (extractor) {
    if (!extractor.chunkSet) {
      extractor.chunkSet = new Set();
    }

    if (extractor.chunkSet.has(chunkName)) {
      return <link />;
    }

    extractor.chunkSet.add(chunkName);

    let cssChunks = extractor.getChunkAssets([chunkName]).filter(v => v.scriptType === 'style');

    if (cssChunks.length > 1) {
      throw new Error(`${chunkName} can only have one css chunk`);
    }

    if (cssChunks.length === 0) {
      return <link />;
    }

    return (
      <link key={cssChunks[0].url} href={cssChunks[0].url} rel="stylesheet" />
    );
  }

  return <link suppressHydrationWarning />;
}
```

theKashey:
`${chunkName} can only have one css chunk` is a wrong direction, as long as it's up to bundler to split.

luojunbin:
Above is just a demo. All I want is to get dynamic css chunks during `renderToNodeStream`.
It seems that I can't get any chunks before `render`, so I want to add these two parameters.

theKashey:
But you really can't do it. `Link` would be rendered during __during__ `renderToNodeStream` and to render links before HTML content you have to interleave stream from react and your own "middleware".
I would recommend not to try to solve your problem this way, please follow the current guidelines. However, if you want to have CSS before HTML during SSR - consider __critical CSS extraction__, and [used-styles](https://github.com/theKashey/used-styles) would help you with it.

PR #487, opened 2019-12-20
perf: avoid calling stats.toJson if possible

hedgepigdaniel:
## Summary

Webpack `stats.toJson` is an expensive operation for large builds, and the loadable-components webpack plugin runs it each emit (so it slows down the watching recompile). Often there are many reasons besides loadable-components to have a stats asset in a build, and large speed savings can be made by combining them into one.

This change avoids running `stats.toJson` (or adding the emit hook at all) if both `outputAsset` and `writeToDisk` are false, in which case the returned value is unused.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -75,7 +75,9 @@ class LoadablePlugin {
     // Add a custom output.jsonpFunction: __LOADABLE_LOADED_CHUNKS__
     compiler.options.output.jsonpFunction = '__LOADABLE_LOADED_CHUNKS__'
 
-    compiler.hooks.emit.tapAsync('@loadable/webpack-plugin', this.handleEmit)
+    if (this.opts.outputAsset || this.opts.writeToDisk) {
+      compiler.hooks.emit.tapAsync('@loadable/webpack-plugin', this.handleEmit)
+    }
   }
 }
 
--- a/website/src/pages/docs/server-side-rendering.mdx
+++ b/website/src/pages/docs/server-side-rendering.mdx
@@ -244,3 +244,25 @@ const extractor = new ChunkExtractor({
   entrypoints: ['client'] // array of webpack entries (default: ['main'])
 })
 ```
+
+## Using your own stats file
+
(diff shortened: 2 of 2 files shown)


theKashey:
changes seem to be correct, but they are bound to the plugin settings. 
So, question, when and why they would be disabled? Should we document this feature if it is expected to be used?

hedgepigdaniel:
I think normally (as in, in a simple idiomatic app with SSR set up like the [SSR guide](https://loadable-components.com/docs/server-side-rendering/)) you wouldn't want to disable `outputAsset`.

It's useful if you have multiple plugins that are calling `stats.toJson` for other reasons, and each of them slowing down recompile time. For example, you might have:
- a plugin to write `stats.json` to disk to synchronize it with some legacy Django setup or similar that still needs to serve `<script />` tags for webpack entry points (anyone else..? lol)
- `webpack-hot-client` needs it
- any plugin that needs the list of chunks/modules or whatever to be available outside the webpack compilation itself - I have seen many, I forgot all of them.

In that case, its best to have just one plugin to call `stats.toJson` and make the result available somehow, and to pass around the result to all the plugins that need it. For example, if you're using `webpack-hot-server-middleware` the stats object is passed to the SSR render function, so in loadable-components you can use the [`stats` option instead of the `statsFile` option](https://loadable-components.com/docs/api-loadable-server/#chunkextractor), and therefore you don't need the webpack plugin to write the stats to disk or to a webpack asset.

In that case it would be possible to not use the webpack plugin at all, except that this bit:

https://github.com/gregberge/loadable-components/blob/853e18b55140bc02e7716a5d34f2b00a6fe9e2c7/packages/webpack-plugin/src/index.js#L75-L76

is necessary for `loadableReady` to work on the client side.

So... I'm not sure what would be best to add to the docs (they document the common/simple case, which is probably a good thing). Maybe just add a note in the SSR guide to point out that its possible to use the ChunkExtractor `stats` option and disable `outputAsset`, and then you can pass through stats object some other way?

hedgepigdaniel:
I've added a section in the SSR guide explaining the issue of `stats.toJson()` being slow and how to avoid it.

PR #506, opened 2020-01-09
fix: safari does not paint the fallback component

hkhere:
While a async chunk is being loaded, safari will not paint the fallback component's dom untill the async chunk is completetly loaded.

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -89,7 +89,9 @@ function createLoadable({ resolve = identity, render, onLoad }) {
         this.mounted = true
 
         if (this.state.loading) {
-          this.loadAsync()
+          setTimeout(() => {
+            this.loadAsync()
+          })
         } else if (!this.state.error) {
           this.triggerOnLoad()
         }
@@ -183,7 +185,15 @@ function createLoadable({ resolve = identity, render, onLoad }) {
 
         if (options.suspense) {
           const cachedResult = this.getCache()
-          if (!cachedResult) throw this.loadAsync()
+          if (!cachedResult) throw new Promise(resolve0 => {
+            setTimeout(() => {
+              setTimeout(() => {
+                this.loadAsync().then(res => {
+                  resolve0(res)
+                })
+              })
+            })
+          })
           return render({
             loading: false,
             fallback: null,


gregberge:
Sorry the code is not correct, could you please add an issue instead to describe your problem.

hkhere:
@gregberge [issue 509](https://github.com/gregberge/loadable-components/issues/509)

theKashey:
Any ideas why your code fixes the problem? Single `setTimeout`, double `setTimeout`... why?

hkhere:
@theKashey The intention is to put the requests of async chunks to other event loop phase, let safari paints the dom of the fallback component first.

It seems safari tries to do some optimizations by delaying the painting of dom if it is requesting some async chunks.  But I did not really examine the source code of webkit.

I expected one `setTimeout ` would do the job but it did not . So I tried double `setTimeout` and it does show the fallback component. Maybe we should/could find an more elegant solution.

theKashey:
I've played with your example a bit, removed all possible side effects like `hot-loader`, even tried different code-splitting libraries, including `React.lazy` - everyone is affected.

Strangely why you were the first one who discovered this bug. 

Wrapping initial load with `timeout` is more or less enough, however, there is no need to wrap `suspense`-related code (and you've broke something with it).

So - to resolve the problem it is enough to delay `css` loading via `setTimeout(fn,1)`, and it's better to implement inside "loadAsync" method, not around it. Unfortunately my favourite "double promise" did not work at all.

Plus - such logic would make `loadable` a bit more stable. For example, right now it could be resolved __almost synchroniously__ if scripts are cached using ServiceWorkers, which might lead to some unpredictable errors.

- [ ] setTimeout should be in `loadAsync`
- [ ] no other code should be affected
- [ ] create a test with _synchroniuos thenable_ (see https://github.com/facebook/react/pull/14626 for more details). It's actually A BIG question how loadable should handle this situation)

theKashey:
It's also enough to call `document.body.offsetLeft` (cause reflow) to solve the issue.

Let's first find a bug in webkit bug tracker.

theKashey:
Ok. I am closing this PR.
Adding any timeouts of another sort of hacks there does not solve the problem:

__Yes__ - you will see "Loading", not __NO__, just after it a __whole application would be frozen until css file is loaded".

I make a simple example, based of your code:
```js
const Home = lazy(() => new Promise(resolve => setTimeout(() => resolve(import('./Home')), 1000)));
// ^^ delay import my 1s to let "loading" be displayed

const App = () => {
  const [state, setState] = useState(0);
  useEffect(
    () => {
      setInterval(() => setState(s => s + 1), 100);
    },
    []
  );
  return <div>
    <React.Suspense fallback={<Loading/>}><Home/></React.Suspense>
    {state}
     here is a counter, changing 10 times per second... 
  </div>
}

export default App;
```
Long story short - counted would not _count_.

This problem should be solved elsewhere.

hkhere:
> Adding any timeouts of another sort of hacks there does not solve the problem

> This problem should be solved elsewhere.

@theKashey  Totally agree.  I should have just opened an issue instead of an pr with terrible code.

theKashey:
@hkhere - you did everything right. You raised the issue, then raised attention to the problem, you led to the underlying problem discovery, and, as a result, you will get your issue fixed.

Issue #508, opened 2020-01-10
Why a separate @loadable/webpack-plugin?


tstirrat15:
## 💬 Questions and Help

Looking at the source for both, [@loadable/webpack-plugin](https://github.com/gregberge/loadable-components/tree/master/packages/webpack-plugin) looks a whole lot like [webpack-stats-plugin](https://github.com/FormidableLabs/webpack-stats-plugin).

One difference between them is that `webpack-stats-plugin` is compatible with the [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer#usage-as-a-cli-utility) CLI, and `@loadable/webpack-plugin` is not. This is down to [`@loadable/webpack-plugin` explicitly excluding parts of the stats file](https://github.com/gregberge/loadable-components/blob/master/packages/webpack-plugin/src/index.js#L19-L27), which the [stats plugin does not](https://github.com/FormidableLabs/webpack-stats-plugin/blob/master/lib/stats-writer-plugin.js#L65-L68).

My questions:

1. Why are those fields excluded at all?
1. If they don't need to be excluded (i.e. adding an option that's passed through to the stats emitter), is there a meaningful difference between this package and `webpack-stats-plugin`?
1. If there isn't, why does this repo reimplement `webpack-stats-plugin` rather than using/peerdepping it?

theKashey:
1. Because loadable don't need them.
2. Yes. They don't need to be excluded
3. To hide the implementation details. And to be concrete - because you don't have to use `webpack-stats-plugin` - it's a ONE call to webpack internals, and you might ask webpack to generate `stat` file for you (`--stats`)

Anyway, the problem with `stats` was "fixed" and documented - https://github.com/gregberge/loadable-components/pull/487

PR #510, opened 2020-01-10
Allow passing of statsOption to toJson method of stats object

tstirrat15:
## Summary
See #508. I didn't do a good job of communicating that my current problem with `@loadable/webpack-plugin` is that it doesn't allow me to use `webpack-bundle-analyzer`, which requires information about chunks and modules, which [this package excludes](https://github.com/gregberge/loadable-components/blob/master/packages/webpack-plugin/src/index.js#L19-L27).

It makes sense to me that this package would default to the bare minimum required to make `@loadable/component` work, but I would also like to be able to pass additional options to the stats dump so that I don't have to dump stats twice (for context, we keep the `stats.json` from each build as an artifact so that we can keep track of changes in bundle size over time).

## Test plan
Add `statsOption: 'normal'` to the plugin options and run a build. See that the resulting stats file contains chunk and module information.

Run the website and see that the option is documented under the API docs for the webpack plugin.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -8,15 +8,7 @@ class LoadablePlugin {
     path,
     writeToDisk,
     outputAsset = true,
-  } = {}) {
-    this.opts = { filename, writeToDisk, outputAsset, path }
-
-    // The Webpack compiler instance
-    this.compiler = null
-  }
-
-  handleEmit = (hookCompiler, callback) => {
-    const stats = hookCompiler.getStats().toJson({
+    statsOption = {
       hash: true,
       publicPath: true,
       assets: true,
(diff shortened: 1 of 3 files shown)


theKashey:
You shall never use `loadable-components` to generate `stats` for `webpack-bundle-analyser`, that's just not right. Keep in mind - that's just a coincidence that the output format of a plugin is compatible with something else.

However, using "full" `stat.json` generated by other plugins seems to be a bit more ok-ish, and this is how it's currently documented.

tstirrat15:
I was under the impression that that wouldn't work because the [webpack plugin injects a jsonp function](https://github.com/gregberge/loadable-components/blob/master/packages/webpack-plugin/src/index.js#L75-L76) used by `loadableReady`. Is that not the case?

As an aside, I was kind of confused by the existence of `loadableReady` at all. I was under the impression that `<script async />` tags preserve the order of execution, even if they download asynchronously/out of order.

tstirrat15:
And for additional context, using `webpack-stats-plugin` to drive SSR + code splitting is how it's done in [react-universal-component](https://github.com/faceyspacey/react-universal-component), which is the library I'm trying to migrate from. It's dead :(

tstirrat15:
I'm completely fine with using `webpack-stats-plugin` and closing this PR if that's a possibility.

I'd also be game to write some documentation for providing `stats.json` through methods other than `@loadable/webpack-plugin` if there's interest in that.

theKashey:
There was a PR by @hedgepigdaniel merged just a few days ago, where you can disable `loadable-stats` generation - https://github.com/gregberge/loadable-components/pull/487
It provides information on how to NOT use `toJson` feature, but not how to USE "stats" generated by other plugins, while I think that's the case.

hedgepigdaniel:
See the new docs: https://loadable-components.com/docs/server-side-rendering/#using-your-own-stats-file

tstirrat15:
Got it. Thank you for the clarification!

PR #511, opened 2020-01-13
Separate delay/timeout/sideload functionality from loadFn

hedgepigdaniel:
## Summary

See https://github.com/gregberge/loadable-components/issues/245 and https://github.com/gregberge/loadable-components/pull/483

Basically the aim is to avoid overloading the `loadFn` so that it is less confusing.

#### New `guard` option
Previously to implement a delay, timeout etc the import promise would need to be wrapped, and the wrapped promise was responsible for resolving to the component. This was confusing because it gave the impression that it was possible to resolve to something else (e.g. a named export), and in fact this worked but only on the client side. The new `guard` option does not have to resolve to anything and its resolved value is consistently ignored, so does not create this confusion:

```typescript
import loadable from '@loadable/component';

const LoadableComponent = loadable(() => import('./xyz'), {
  guard: (importPromise) => pMinDelay(importPromise, 1000).then(() => {}),
});
```

####  `guard` has access to props
```typescript
import loadable from '@loadable/component';
const DynamicLoadableComponent = loadable(() => import('./xyz'), {
  guard: (importPromise, props) => pMinDelay(importPromise, props.timeout).then(() => {}),
});
```

## Detailed choices

- The babel plugin has been modified to throw an error if the import creator does anything except directly returning an `import()` statement. This is not necessary for this feature to work (e.g. the error could be hidden behind a flag, or generate a warning now and and an error later). I think it's a good idea though, because it removes the trap where wrapping the `import()` statement (e.g. `loadable(() => import('./xyz').then((mod) => mod.Component))`) works on the client side but not the server side
- To replace the practice of implementing delays/timeouts by wrapping the import promise (e.g. as in the docs for [delay](https://loadable-components.com/docs/delay/)), a new flag is added called `guard`. The consequences of using it are much the same, but this allows the babel plugin to enforce that the `import()` promise is never wrapped (therefore avoiding the client vs SSR confusion). Again, not necessary for the feature to work, but a good idea IMO.

## Test plan

Existing tests have been changed where necessary and new tests have been added for new functionality

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -580,62 +580,6 @@ exports[`plugin loadable.lib should be transpiled too 1`] = `
 });"
 `;
 
-exports[`plugin simple import in a complex promise should work 1`] = `
-"loadable({
-  resolved: {},
-
-  chunkName() {
-    return \\"ModA\\";
-  },
-
-  isReady(props) {
-    const key = this.resolve(props);
-
-    if (this.resolved[key] === false) {
-      return false;
-    }
-
-    if (typeof __webpack_modules__ !== 'undefined') {
-      return !!__webpack_modules__[key];
-    }
-
-    return false;
-  },
-
-  importAsync: () => timeout(import(
-  /* webpackChunkName: \\"ModA\\" */
-  './ModA'), 2000),
-
-  requireAsync(props) {
-    const key = this.resolve(props);
-    this.resolved[key] = false;
-    return this.importAsync(props).then(resolved => {
-      this.resolved[key] = true;
-      return resolved;
-    });
-  },
-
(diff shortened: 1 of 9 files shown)


hedgepigdaniel:
perhaps `guard` should be called `clientLoadExtra`,  so its 100% clear that it doesn't work on the server side

theKashey:
- by a some reason I am thinking about `driver` like `asyncDriver`. 
- or something very simple, like `onAsync`/`ifAsync`, as long as it takes a place "on" "async"
- or not `async`, but `lazy`. Because it works when component is lazy, not eager loaded.

hedgepigdaniel:
bot pls, discussion still in progress

hedgepigdaniel:
bump

theKashey:
Let's keep it open

theKashey:
🧟

PR #513, opened 2020-01-14
docs: fix ChunkExtractor.outputPath comment

oreqizer:
## Summary

The `ChunkExtractor.outputPath` option is also used by functions like `getCssString`, not only `requireEntrypoint`, so I adjusted that.


gregberge:
Thanks!

PR #520, opened 2020-02-03
Fix crashing with babel-plugin-dynamic-import-node-sync

nikityy:
## Summary
When both `@loadable/babel-plugin` and `babel-plugin-dynamic-import-node-sync` are defined in plugins, `@loadable/component` crashes on server complaining that `.then is not callable`. Seems like this happens because `babel-plugin-dynamic-import-node-sync` does its job before `@loadable/component`, and `importAsync` function's body becomes just `require(PATH)`, which just returns the module, not a promise.

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -31,7 +31,7 @@ exports[`plugin Magic comment should remove only needed comments 1`] = `
   requireAsync(props) {
     const key = this.resolve(props);
     this.resolved[key] = false;
-    return this.importAsync(props).then(resolved => {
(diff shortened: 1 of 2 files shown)


gregberge:
You should not use `babel-plugin-dynamic-import-node-sync` why would you use it?

nikityy:
> You should not use babel-plugin-dynamic-import-node-sync why would you use it?

`loadable-components` seems to be responsible just for React components, not all dynamic imports in the app.

theKashey:
The change itself is ok. But if `importAsync` is expected to return a Promise - it shall return a Promise. You should change it, not the place where it is used.

gregberge:
@nikityy can you wrap your Loadable in a Promise.resolve in your project instead of for everyone?

PR #521, opened 2020-02-04
Fixed lazy usage with Suspense and Error Boundary together

kamikazePT:
## Summary

When using lazy ( the Suspense version of loadable ), if any error occured, the Suspense would keep rendering It's fallback and retrying and never being catched by the parent ErrorBoundary.

For example I have a dynamic Icon that has specific components, and depending on the Icon type, it loads the correct component, displaying a default icon if no component is found, the way to do this without an Error Boundary would be to add the catch clause on the import( ) statement.

## Test plan

With this PR, if the import( ) fails because the file doesn't exist, it will then render the Fallback provided to the ErrorBoundary

Screenshot from 2020-02-04 21-07-39

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-use-before-define, react/no-multi-comp, no-underscore-dangle */
 import React from 'react'
-import { invariant, statusAware, STATUS_PENDING, STATUS_REJECTED } from './util'
+import { invariant, statusAware, STATUS_REJECTED } from './util'
 import Context from './Context'
 
 function resolveConstructor(ctor) {
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -186,7 +186,7 @@ function createLoadable({ resolve = identity, render, onLoad }) {
         } = this.props
         const { error, loading, result } = this.state
 
-        if (options.suspense) {
+        if (options.suspense && !this.promise) {
           const cachedResponse = this.getCache()
           if (!cachedResponse) throw this.loadAsync()
           if (cachedResponse.error) throw cachedResponse.error
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -193,10 +193,8 @@ function createLoadable({ resolve = identity, render, onLoad }) {
 
         if (options.suspense) {
           const cachedPromise = this.getCache()
-          if (!cachedPromise) {
+          if (!cachedPromise || cachedPromise.status === STATUS_PENDING) {
             throw this.loadAsync()
-          } else if (cachedPromise.status === STATUS_PENDING) {
-            throw cachedPromise
           }
         }
 
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -129,13 +129,8 @@ function createLoadable({
             error: undefined,
             loading: true,
           })
-
-          if (!options.suspense) this.loadAsync()
-        } else if (this.state.loading) {
-          this.loadAsync()
-        } else if (!this.state.error) {
-          this.triggerOnLoad()
         }
+        this.loadAsync()
       }
 
       componentDidUpdate(prevProps, prevState) {
@@ -247,8 +242,6 @@ function createLoadable({
         }
 
         return render({
-          loading,
-          fallback,
           result,
           options,
           props: { ...props, ref: forwardedRef },
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -151,25 +151,25 @@ function createLoadable({ resolve = identity, render, onLoad }) {
       loadAsync() {
         if (!this.promise) {
           const { __chunkExtractor, forwardedRef, ...props } = this.props
-          this.promise = ctor
-            .requireAsync(props)
+
+          const cachedPromise = this.getCache()
+
+          this.promise = cachedPromise || ctor.requireAsync(props)
+
+          if (options.suspense) this.setCache(this.promise)
+
+          this.promise = this.promise
             .then(loadedModule => {
               const result = resolve(loadedModule, { Loadable })
-              if (options.suspense) {
-                this.setCache({ result })
-              }
               this.safeSetState(
                 {
-                  result: resolve(loadedModule, { Loadable }),
+                  result,
                   loading: false,
                 },
                 () => this.triggerOnLoad(),
               )
             })
             .catch(error => {
-              if (options.suspense) {
-                this.setCache({ error })
-              }
               this.safeSetState({ error, loading: false })
             })
         }
@@ -186,18 +186,9 @@ function createLoadable({ resolve = identity, render, onLoad }) {
         } = this.props
         const { error, loading, result } = this.state
 
-        if (options.suspense && !this.promise) {
-          const cachedResponse = this.getCache()
-          if (!cachedResponse) throw this.loadAsync()
-          if (cachedResponse.error) throw cachedResponse.error
-
-          return render({
-            loading: false,
-            fallback: null,
-            result: cachedResponse.result,
-            options,
-            props: { ...props, ref: forwardedRef },
-          })
+        if (options.suspense) {
+          const cachedPromise = this.getCache()
+          if (!cachedPromise) throw this.loadAsync()
         }
 
         if (error) {
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -189,7 +189,7 @@ function createLoadable({ resolve = identity, render, onLoad }) {
         if (options.suspense) {
           const cachedResponse = this.getCache()
           if (!cachedResponse) throw this.loadAsync()
-          if (!cachedResponse.error) throw cachedResponse.error
+          if (cachedResponse.error) throw cachedResponse.error
 
           return render({
             loading: false,
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -130,13 +130,13 @@ function createLoadable({
             loading: true,
           })
         }
-        this.loadAsync()
+        this.loadAsyncOnLifecycle()
       }
 
       componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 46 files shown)


kamikazePT:
In the meantime I've found another issue regarding Suspense usage.

If I have a full dynamic import aka something like
` lazy(props => import(`./${props.name}`))`

and "name" is changing runtime, we would be blocked on a loop on the throw loadAsync call

kamikazePT:
@gregberge yeah... but caching the promise, how would you get the result in a sync way to change this piece of code inside the render method? to cache the promise instead of the result/error the render method needs to be re-thinked, no?

```
      if (options.suspense && !this.promise) {
          const cachedResponse = this.getCache()
          if (!cachedResponse) throw this.loadAsync()
          if (cachedResponse.error) throw cachedResponse.error

          return render({
            loading: false,
            fallback: null,
            result: cachedResponse.result,
            options,
            props: { ...props, ref: forwardedRef },
          })
        }

```

kamikazePT:
@gregberge 
pushed changes :)

tests are passing, promise cached and render method looks cleaner

theKashey:
There is another PR, which fixes the same problem - https://github.com/gregberge/loadable-components/pull/516

kamikazePT:
Actually I think that PR is somewhat a complement to this one, he doesnt fix the error not being propagated to the error boundary, he fix's the re-fetching on rehydration after ssr

Atleast that was what I understood from the change he made which was on loadSync that runs on ssr

Still I dont believe that the cache persists from ssr to csr o.O which makes his fix do nothing

packages/component/src/createLoadable.js
```diff
@@ -184,15 +187,8 @@ function createLoadable({ resolve = identity, render, onLoad }) {
...
-          if (!cachedResult) throw this.loadAsync()
-          return render({
-            loading: false,
-            fallback: null,
-            result: cachedResult,
-            options,
-            props: { ...props, ref: forwardedRef },
-          })
+          const cachedPromise = this.getCache()
+          if (!cachedPromise) throw this.loadAsync()
```
theKashey: That would happen if you have two components? What if this component would be rendered twice?
In both cases the second one would be  "cached" and probably would not work.

Probably the cache should hold information about the status of a cached item.

packages/component/src/createLoadable.js
```diff
@@ -151,16 +151,19 @@ function createLoadable({ resolve = identity, render, onLoad }) {
       loadAsync() {
         if (!this.promise) {
           const { __chunkExtractor, forwardedRef, ...props } = this.props
-          this.promise = ctor
-            .requireAsync(props)
+
+          const cachedPromise = this.getCache()
+
+          this.promise = cachedPromise || ctor.requireAsync(props)
```
theKashey: What if the cache is a bit obsolete?
What if it was rejected by any reason? 

Now you are not able to restart the operation.

packages/component/src/loadable.test.js
```diff
@@ -242,6 +279,21 @@ describe('#lazy.lib', () => {
...
+    const { container } = render(
+      <ErrorBoundary fallback="error">
+        <React.Suspense fallback="progress">
+          <Lib>{renderFn}</Lib>
+        </React.Suspense>
+      </ErrorBoundary>,
+    )
+    expect(container).toHaveTextContent('progress')
+    load.reject(new Error('Error Boundary'))
+    await wait(() => expect(container).toHaveTextContent('error'))
```
theKashey: Please add one more test - `supports retry from the Error Boundary`. It shall not be a final, unrecoverable, state.

packages/component/src/createLoadable.js
kamikazePT: If you have 2 components at the same tree level, they should have different Keys (like in arrays), having different Keys makes them different instances with their own internal cache

What are your thoughts on that?

packages/component/src/createLoadable.js
kamikazePT: I would make the promise retryable, instead of re-rendering from the error boundary :) although there might be use cases for the re-render, I'll think for a bit about this

packages/component/src/createLoadable.js
theKashey: You might code split  absolutely everything, including a single `Button`, you can use 20 times on a single page.
So - it could be the same cache key, repeated 20 times.

packages/component/src/createLoadable.js
theKashey: Promise cannot be retryable once it resolved or rejected - those are final states.
And in case of some error the normal expectation nowadays is to throw that error, and handle it at the nearest boundary.

kamikazePT: this was what I had in mind when I said "making" the promise retryable
```
retry = (fn, retries = 3) =>
    fn().catch(err => 
            retries > 1 
                ? retry(fn, retries - 1) 
                : Promise.reject(err))

loadable(() => retry(() => import('foo')))
```

packages/component/src/createLoadable.js
kamikazePT: I've added some tests for multiple rendered components
Please check them out to see if I understood you correctly

gregberge:
Hello @kamikazePT, thank you for trying, but I think the approach is not the good one. The best one is to create a container that encapsulate the promise.

An example:

```js
export const PENDING = 0
export const RESOLVED = 1
export const REJECTED = 2

function createContainer(status, result) {
  const container = {
    status,
    result,
    write(result) {
      if (container.status !== RESOLVED) {
        throw new Error('Cannot write container if not resolved')
      }
      return { ...container, result, toJSON: () => result }
    },
    toJSON() {
      return result
    },
  }
  return container
}

export function createContainerFromPromise(promise) {
  const container = createContainer(PENDING)
  container.promise = promise.then(
    r => {
      container.status = RESOLVED
      container.result = r
      container.toJSON = () => r
    },
    e => {
      container.status = REJECTED
      container.error = e
    },
  )
  return container
}

export function createContainerFromValue(value) {
  return createContainer(RESOLVED, value)
}
```

packages/component/src/createLoadable.js
theKashey: It's good to have "retry" working out of the box, but usually, it's better to expose this ability to the user, as long as they might want to communicate the problem to the end-user.

kamikazePT:
@gregberge 
@theKashey 

So with some hammer magic I managed to make retrying work for both lazy and loadable variants

It needs some code cleaning but atleast it's tested and working at the moment...

packages/component/src/util.js
```diff
@@ -12,3 +12,33 @@ export function warn(message) {
...
 }
+
+export const STATUS_PENDING = 'PENDING'
+export const STATUS_RESOLVED = 'RESOLVED'
+export const STATUS_REJECTED = 'REJECTED'
+
+export function statusAware(promise) {
+  let status = STATUS_PENDING
+
+  const proxy = new Proxy(promise, {
```
theKashey: No `Proxies` please. They are not supported widely enough yet.
And why this needs a recursive wrapping?

packages/component/src/createLoadable.js
```diff
@@ -90,7 +90,17 @@ function createLoadable({ resolve = identity, render, onLoad }) {
...
+        const cachedPromise = this.getCache()
+
+        if (cachedPromise && cachedPromise.status === STATUS_REJECTED) {
+          this.setCache()
+          this.setState({
+            error: undefined,
+            loading: true,
+          })
+
+          if (!options.suspense) this.loadAsync()
```
theKashey: You can load it async with suspense as well, why not?

packages/component/src/util.js
kamikazePT: The recursion is that if you call then or catch, it enables you to access status on that newly created promises from those calls

packages/component/src/createLoadable.js
kamikazePT: Because you need to throw the loadAsync call on the render method so that Suspense catches it, doing that on DidMount makes you add a lot more if conditions at the render method to support more corner cases....sometimes the Best Solution to a problem is to avoid that problem :/

kamikazePT: @theKashey 
Besides adding a fallback for Proxy, is there anything else preventing the acceptance of this PR?

kamikazePT:
@theKashey 
@gregberge 

Besides adding a fallback for Proxy usage when not available, is there anything else that would block your acceptance?
I have been busy but I'll proceed with that change in a few days

gregberge:
@kamikazePT I am sorry but i don't have time to dig into this right now.

kamikazePT:
using a fork with this PR in a production app, just noticed I added a bug when fixed the retrying functionality

also tests weren't exactly bulletproof and could lead to false assumptions, they were improved as well

theKashey:
Let's double-check what's missing.

kamikazePT:
@theKashey 
PR updated with latest commits on master

theKashey:
👀

kamikazePT:
@theKashey 

revisited the PR, removed the ES6 Proxy

kamikazePT:
@theKashey 
@gregberge 

Anything you need to discuss regarding the solution? or the refactor in the unit tests mocks of the load function?

theKashey:
Hey @kamikazePT, sorry for being so slow. Just two moments left:
- [ ] `::` operator used in test (is it "babel-bind"? Is there any way not to use it?)
- [ ] adding callbacks to pending promises

kamikazePT:
@theKashey 

the bind operator is only there for sugar syntax and since it's only being used in testing environment, it won't add overhead to the bundling / transpiling for the release files,  in my honest opinion I would keep it :)

can you explain a bit more what you mean with adding callbacks for pending promises?

theKashey:
> can you explain a bit more what you mean with adding callbacks for pending promises?

Every time you call `.loadAsync` it does `.then` It is possible to call it multiple times, due to multiple reasons, and it will "stack" callbacks, while only one is required.

kamikazePT:
@theKashey 

Fixed the callback situation and removed unnecessary bits of code that weren't doing anything (tests are passing)

Please check it again :)
Hopefully everything is sorted!

theKashey:
Very good @kamikazePT 👍

theKashey:
PS: 🥳

Issue #337, opened 2019-05-15
Add the choose between defer or async


gregberge:
## ✨Enhancement

Add the ability to insert scripts using defer or async. By default, async is the best choice but it could creates some issues with CSS.

See #258 for context.

yairkukielka:
Thanks for opening this issue! 

Actually, it's not only css problems, but js too. See [this](https://stackoverflow.com/a/39711009/3032209) stackoverflow answer for more context. 

There's 3 options here: 

- <script>
- <script async>
- <script defer>


If you scripts are independent from each other using `async` is totally fine (and even recommended for speed).

If you have a script that depends on another script `async` doesn't guarantee that they will load in the right order, meaning that they will fail at runtime depending on the browsers implementation. You'll have to either load them synchronously (no attribute) or use `defer` if you want your scripts to work in a deterministic way.

neoziro:
@yairkukielka thanks for the precision! We have to address that point seriously.

nathanjd:
Thanks, @yairkukielka for the list. I have a few additions.

The way async and defer behave is different depending on whether the script is dynamically injected or present during the first html pass. For example, to get "deferred" behavior out of a dynamically inject script, it must use `<script async="false"></script>` instead of the deferred attribute. To preserve consistent behavior, chunks extracted for server render would have to rendered with `<script defer></script>` while those same chunks would need `<script async="false"></script>` when injected through the loadable() runtime.

Async:
> Dynamically inserted scripts (using document.createElement()) load asynchronously by default, so to turn on synchronous loading (i.e. scripts load in the order they were inserted) set async="false".

Defer:
> To achieve a similar effect for dynamically inserted scripts use async="false" instead. Scripts with the defer attribute will execute in the order in which they appear in the document.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Attributes

Personally I think it's best to leave resource hints such as `<link rel="prefetch">` out of this feature request since they don't control execution of scripts, just network prefetch.

theKashey:
Async is quite interesting, as long as it uses CPU while Network is still in use, not interleaving their work. 
However - the result is not quite “safe” - script could be ready far prior DOM, and you have to wrap boot with old good DOM Ready.

What would be really interesting - <link rel=“modulepreload”> - it will do everything off main thread. 
https://philipwalton.com/articles/using-native-javascript-modules-in-production-today/
https://developers.google.com/web/updates/2017/12/modulepreload
[...]

PR #526, opened 2020-02-12
improvement(server): Allow setting async/defer via extraProps

jackyef:
## Summary

Solving https://github.com/gregberge/loadable-components/issues/337

Default: `async: true, defer: false`

Scenarios: 
* `extractor.getScriptTags({ defer: true })` will return `<script async defer ...></script>` (because `async` is `true` by default)
* `extractor.getScriptTags({ async: true })` will return `<script async ...></script>`
* `extractor.getScriptTags({ async: false, defer: true })` will return `<script defer ...></script>`
* `extractor.getScriptTags()` will return `<script async ...></script>` to keep the default behaviour, preventing breaking change.

Using both `async` and `defer` together might have a valid use case as explained [here](https://stackoverflow.com/questions/50615101/what-does-async-defer-do-when-used-together)

## Test plan

Tests are added to `ChunkExtractor.test.js`

### DIFF
--- a/packages/component/.size-snapshot.json
+++ b/packages/component/.size-snapshot.json
@@ -1,0 +1,21 @@
+{
+  "dist/loadable.cjs.js": {
+    "bundled": 12852,
+    "minified": 6137,
+    "gzipped": 2208
+  },
+  "dist/loadable.esm.js": {
+    "bundled": 12469,
+    "minified": 5828,
+    "gzipped": 2141,
+    "treeshaked": {
+      "rollup": {
+        "code": 259,
+        "import_statements": 259
+      },
+      "webpack": {
+        "code": 5061
+      }
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


jackyef:
> The method getRequiredChunksScriptTag should support async and defer like other

One question about this though, does `async`/`defer` has impact when the script is inline? (Which is the case of `getRequiredChunksScriptTag`)

jackyef:
@gregberge Hi, are we really deciding on this as `wontfix`?

ArgonAlex:
Just want to express my desire for this functionality. Right now I'm resorting to hackily string replacing ` async ` with ` defer `.

suhanw:
Would love this feature too

theKashey:
Stalebot go away

yukiyang0729:
Would love this feature too！！

yukiyang0729:
Would love this feature too！！

Issue #538, opened 2020-03-14
@loadable/babel-plugin not support webpack magic comments


kaysonwu:
## 🐛 Bug Report
`@loadable/babel-plugin`、`Module not found`

As long as I remove `@loadable/babel-plugin`  from` babel.config.json`  it compiles correctly.  

```bash
ERROR in ./src/components/router/index.tsx
Module not found: Error: Can't resolve '@/pages/' in 'D:\htdocs\react\react-app\src\components\router'
 @ ./src/components/router/index.tsx 54:13-44 26:11-30:15
 @ ./src/components/application/index.tsx
 @ ./src/server.tsx
```

The code reporting the error is:

```typescript
const Page = loadable((props: any) => import(/* webpackChunkName: "pages/[request]" */`@/pages/${props.path}`), {
  cacheKey: props => props.path
});
```
 I configured `resolve.alias`  in `webpack`,  The error persists even though I use relative path.

## To Reproduce

 [react-app](https://github.com/kaysonwu/react-app/tree/3.x).

```bash
git clone -b 3.x https://github.com/kaysonwu/react-app.git

yarn install && yarn build:server
```

## Expected behavior

Compile correctly

## Link to repl or repo (highly encouraged)

 [react-app](https://github.com/kaysonwu/react-app/tree/3.x).

```bash
## System:
 - OS: Windows 10 10.0.18363
 - CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
 - Memory: 1.74 GB / 7.89 GB
## Binaries:
 - Node: 12.13.1
 - Yarn: 1.21.1
 - npm: 6.13.4
## npmPackages:
 - @loadable/babel-plugin: ^5.12.0 => 5.12.0 
 - @loadable/component: ^5.12.0 => 5.12.0
 - @loadable/server: ^5.12.0 => 5.12.0
 - @loadable/webpack-plugin: ^5.12.0 => 5.12.0
```

kaysonwu:
@gregberge

theKashey:
For now - just don't use module aliases.
Plus could you explain how you configured them.

kaysonwu:
@theKashey  I have provided the project, you can download it and try it. Unfortunately, even with relative paths, errors are reported.

theKashey:
out of curiosity (sorry, I cannot download and try your example right now) - try `'@/pages/'+props.path`, ie a bit simpler construction.
[...]

PR #541, opened 2020-03-23
Fix windows env of SSR example

kaysonwu:
Compilation can not be completed successfully under windows system, it need to use the `cross-env` library

### DIFF
--- a/examples/server-side-rendering/package.json
+++ b/examples/server-side-rendering/package.json
@@ -2,10 +2,10 @@
   "private": true,
   "scripts": {
     "dev": "nodemon src/server/main.js",
-    "build": "NODE_ENV=production yarn build:webpack && yarn build:lib",
+    "build": "cross-env NODE_ENV=production yarn build:webpack && yarn build:lib",
     "build:webpack": "webpack",
     "build:lib": "babel -d lib src",
-    "start": "NODE_ENV=production node lib/server/main.js"
+    "start": "cross-env NODE_ENV=production node lib/server/main.js"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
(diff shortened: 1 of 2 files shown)


kaysonwu:
Fix ts-loader cannot find module after compilation PR please see [#538](https://github.com/gregberge/loadable-components/issues/538)

examples/server-side-rendering/package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "core-js": "^3.0.1",
+    "cross-env": "^7.0.0",
```
theKashey: devDependency please.

packages/babel-plugin/src/properties/chunkName.js
```diff
@@ -132,7 +132,6 @@ export default function chunkNameProperty({ types: t }) {
     const values = getExistingChunkNameComment(callPath)
 
     if (!agressiveImport && values) {
-      addOrReplaceChunkNameComment(callPath, values)
```
theKashey: this change should be separated from `crossEnv`, and have a clear test to prove it's necessity  and correctness.

examples/server-side-rendering/package.json
kaysonwu: The `start` script command also use `NODE_ENV`

packages/babel-plugin/src/properties/chunkName.js
kaysonwu: This seems to be github automatic behavior. I have run the tests locally

kaysonwu:
split PR

examples/server-side-rendering/package.json
theKashey: all scripts are "dev" scripts. Final users are going to use `/dist` + `deps`. Everything not required for the final users should be gone.

PR #542, opened 2020-03-23
Fix ts-loader cannot find module after compilation

kaysonwu:
## Summary

After testing, I have clearly identified the source of the problem. Please refer to my minimal reproduction [loadable-babel-plugin](https://github.com/kaysonwu/minimal-reproduction/tree/master/loadable-babel-plugin)

If you just compile with Babel, `Page` is an aggressive import. Unfortunately, when compiled with `ts-loader` in` webpack`, `Page` becomes` BinaryExpression` type.

Personally, the responsibility of the `chunkName` property is to supplement the` webpackChunkName` annotation for import. So if you set `webpackChunkName` yourself, it should not do any work. Therefore, I deleted this line of [code](https://github.com/gregberge/loadable-components/blob/master/packages/babel-plugin/src/properties/chunkName.js#L135)

## Test plan

test

### DIFF
--- a/packages/babel-plugin/src/properties/chunkName.js
+++ b/packages/babel-plugin/src/properties/chunkName.js
@@ -132,7 +132,6 @@ export default function chunkNameProperty({ types: t }) {
     const values = getExistingChunkNameComment(callPath)
 
     if (!agressiveImport && values) {
-      addOrReplaceChunkNameComment(callPath, values)
       return t.stringLiteral(values.webpackChunkName)
     }
 


kaysonwu:
Fix [#538](https://github.com/gregberge/loadable-components/issues/538)

theKashey:
The fix could not accepted without a prove that something was wrong.
That means that you should have:
- a test, red before, and green now.
- or a test "wrong" before, and "corrected" now.

Please understand this correctly - there is no way we could release a new version package for 600k monthly users, without knowing that we are doing a right thing.

kaysonwu:
@theKashey Please refer to my minimal reproduction [loadable-babel-plugin](https://github.com/kaysonwu/minimal-reproduction/tree/master/loadable-babel-plugin). This problem will only appear after compiling through `ts-loader`

You will find errors by running `yarn build` and `yarn build:babel`

theKashey:
Yep. I've just checked the code - there are no real tests for this moment, that's why you could pass them. 
But you are actually going to break so many things.

kaysonwu:
Ah! If there is an impact, the original test will not pass. However, I can try to add a webback test to prove the bug

PR #544, opened 2020-03-25
Use make-dir instead of mkdirp

pustovalov:
## Summary

```
warning @loadable/webpack-plugin > mkdirp@0.5.4: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
```
https://github.com/isaacs/node-mkdirp/blob/master/CHANGELOG.md

also

### DIFF
--- a/packages/webpack-plugin/package.json
+++ b/packages/webpack-plugin/package.json
@@ -28,6 +28,6 @@
     "webpack": ">=4.6.0"
   },
   "dependencies": {
-    "mkdirp": "^0.5.1"
+    "make-dir": "^3.0.2"
   }
 }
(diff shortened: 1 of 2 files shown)


theKashey:
Ok, so updated `mkdirp` requires `node>=10`, and we have to follow.

pustovalov:
Found one more inconsistency, webpack 4 supports `"node": ">=6.11.5"`
https://github.com/webpack/webpack/blob/webpack-4/package.json#L88

I see a few options:

1. Update mkdirp and drop node 8
2. Switch to another package: https://github.com/sindresorhus/make-dir
like babel did:
https://github.com/babel/babel-loader/commit/47958ca3857e489e450b27c6f3266375015cba82

theKashey:
The second option sounds better, and the `make-dir` library is actually doing only what we need, not a dozen other things.

gregberge:
Thanks!

PR #551, opened 2020-04-11
fix custom webpackChunkName for aggressive import

kaysonwu:
## Summary

`babel-plugin` will ignore the correct custom `webpackChunkName` during dynamic import.  I think it makes sense to keep this, it can store chunks in the specified directory.

## Test plan

error

success


kaysonwu:
Waiting for repair check

PR #567, opened 2020-05-13
fix: memory leak in module cache management, fixes #560

### LINKED ISSUE #560: Memory leak caused by ChunkExtractor
## 🐛 Bug Report

Server side rendering in combination with specifying the stats file path for the ChunkExtractor seems to cause a memory leak. It might be in connection with streaming the response, but I'm not sure about this yet. Lucky wise I found a better solution and think we should adapt the documentation to preload the file only once anyway. Whats the point of loading it for every request again? Do you agree and maybe know the issue of the memory leak?

## To Reproduce

Example for Fastify:
```
const chunkExtractor = new ChunkExtractor({ statsFile: path });
  const jsx = await createReactTree(chunkExtractor);
  const resStream = new PassThrough();
  reply
    .headers({})
    .status(200)
    .send(resStream);
  resStream.write(buildHeader(chunkExtractor));
  const stream = renderToNodeStream(jsx);
  await new Promise((resolve, reject) => {
    stream.pipe(resStream, { end: false });
    stream.on("error", err => {
      resStream.end();
      reject(err);
    });
    stream.on("end", () => {
      resStream.write(buildFooter(client));
      resStream.end();
      resolve();
    });
  });
```
## Expected behavior

ChunkExtractor loads stats and releases all allocated memory after response is completed.

## Used solution

I was able to solve it in my case by simply loading the file once during startup. If the file otherwise gets loaded for every request, then the examples and documentation should also be changed, as it will decrease performance and doesn't seem to be necessary:

```
const statsFile = resolve(__dirname, "loadable-stats.json");
const stats = JSON.parse(readFileSync(statsFile).toString("utf8"));

const doAnswer = async (req, res) => {
   const chunkExtractor = new ChunkExtractor({ stats });
   // more code
}
```
## Link to repl or repo

Unfortunately, this is private project. But as I already found a better solution, I just like to inform you in case you might be aware of the underlying issue.

## System:
 - OS: Linux 5.5 Fedora 31 (Workstation Edition) 31 (Workstation Edition)
 - CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
 - Memory: 23.52 GB / 31.08 GB
 - Container: Yes
 - Shell: 5.0.11 - /bin/bash
## Binaries:
 - Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
 - Yarn: 1.22.4 - /usr/bin/yarn
 - npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
## npmPackages:
 - @loadable/babel-plugin: ^5.11.0 => 5.12.0 
 - @loadable/component: ^5.11.0 => 5.12.0 
 - @loadable/server: ^5.11.0 => 5.12.0 
 - @loadable/webpack-plugin: ^5.7.1 => 5.12.0

theKashey:
## Summary

`smartRequire` is cleaning the cache before requiring module, and that might keep reference to an older version, causing a memory leak.

## Fix

- remove module from the cache
- (added) remove module from parent.children
- (added) remove module from children.parent

## Test plan

The algorithm is borrowed from [rewiremock](https://github.com/theKashey/rewiremock/blob/master/src/wipeCache.js#L34)

### DIFF
--- a/packages/server/src/util.js
+++ b/packages/server/src/util.js
@@ -1,4 +1,21 @@
-export const clearModuleCache = key => delete require.cache[key]
+export const clearModuleCache = (moduleName) => {
+  const m = require.cache[moduleName];
+  if (m) {
+    // remove self from own parents
+    if (m.parent && m.parent.children) {
+      m.parent.children = m.parent.children.filter(x => x !== m);
+    }
+    // remove self from own children
+    if (m.children) {
+      m.children.forEach(child => {
+        if (child.parent && child.parent === m) {
+          child.parent = null;
+        }
+      })
+    }
+    delete require.cache[moduleName]
+  }
+}
 
 export const smartRequire = modulePath => {
   if (process.env.NODE_ENV !== 'production') {


ianvieira:
@theKashey When will this fixed be released? We are having this issue too but it seems it was not released yet. Do we have an ETA?

theKashey:
I will try to arrange a new release this weekends. There is one more important PR to merge before everything.

PR #568, opened 2020-05-14
Allow webpack cache isReady only for initial chunks

theKashey:
## Summary

Only SSR-ed chunks could be tested for `isReady` via webpack cache. Others should go through `requireAsync` and track their process in a internal state variables.


theKashey:
Tested, and it's working 👍 
However, I've found an issue with `loadable.lib` and `moment` - it's actually not working in production mode. More like webpack/moment related problem like moment is `moment` node, and `{default:moment}` in the browser. Result - 💩

PR #573, opened 2020-05-19
feat(esm): ssr compliant esm modules (#1)

marcmrf:
## Summary

The purpose of this PR is to discuss how to support native esm modules, via `webpackIgnore` flag, which opts out code-splitting for that specific module, using native import instead for loading the file.

The approach I followed has been to differentiate two kind of loadable modules, webpack chunks and esm ones, applying different loading strategies for each one.

Loadable components strongly rely in webpack for bypassing hydration limitations for async components, generating a manifest of chunks to preload before being able to render the main app. Once the chunks are loaded, webpack `require.resolveWeak` does the magic for loading synchronously the modules that have been preloaded before.

`webpackIgnore` magic comment completely opts out the module to import out of webpack, having webpack 0 knowledge of it in the generated stats. As a consequence, the module will not be present in the manifest generated by loadable-components webpack plugin either.

In order to being able to ssr properly the components and hydrate them after in the browser, the approach I decided to follow is to apply partial hydration using [react-prerendered-component](https://github.com/theKashey/react-prerendered-component) package. The idea consists in not hydrating these esm modules on the first render, doing it when the component is downloaded instead. So, esm modules will not be present in the array of mandatory modules to load before bootstrapping the app. The html sent by the server will be used instead for the hydration.

Developers are not going to be aware of `react-prerendered-component`. Loadable components will decide itself whether to add it or not depending on the kind of module we're working with.

On the other hand, the native import is going to be replaced by a normal require if the target is node, so the module can be properly imported synchronously.

## Future steps
* Allow to apply progressive hydration in webpack chunks, which will allow users to optimise their apps. Most of the times the content needs to be part of the critical path, but to hydrating everything at the beginning is generally not the most optimal. Developers can create chunks for two reasons, dynamic modules or performance. In case of performance, these modules are created for not affecting the bootstrapping of the app, being the feature they implement not necessary to interact with at the beginning. To download all the involved chunks at the beginning goes against what the developer wants to do most of the time.
* Allow to preload native modules, so they can be load faster when it's their time to render.

## Other solutions considered

Right now loading components delegates to the critical chunk the responsibility of notifying it has been loaded so `require.resolveWeak` can be used. Moving the chunk/module preloading to the main app can solve the issue too, together with a components cache which will return the module synchronously if it's already loaded. This solution seems to introduce more breaking changes.

## Test plan

* Compile all the packages in loadable-components repo
* Link them in server-side-rendering example
* `yarn build`
* `yarn start`
* check the html returned by the server, the component has to be rendered
* click on click me button, the counter is going to be increased.


examples/server-side-rendering/babel.config.js
```diff
@@ -23,6 +27,9 @@ module.exports = api => {
         },
       ],
     ],
-    plugins: ['@babel/plugin-syntax-dynamic-import', '@loadable/babel-plugin'],
+    plugins: [
+      '@babel/plugin-syntax-dynamic-import',
+      ['@loadable/babel-plugin', { target: api.caller(getTarget) }]
```
marcmrf, line 32: the plugin needs to now if the target is node or browser environment, so the import method to use is require or native esm import

packages/component/rollup.config.js
```diff
@@ -45,7 +45,9 @@ export default [
     external: Object.keys(globals),
     plugins: [
       babel(getBabelOptions({ useESModules: true })),
-      nodeResolve(),
+      nodeResolve({
+         mainFields: ['main'],
```
marcmrf, line 49: build was triggering error because of something wrong in react-prerendered-component module version. I can invest some time fixing it if you are interested in the feature

theKashey:
That is a very big PR which solves a problem which might be not a problem in the future.
Upcoming React version gonna support suspense-based `partial hydration` out of the box, where you will be able to load `esm` modules with a simple `React.lazy`.

The reason to use loadable is to _track_ what you've used during SSR, to be able to properly rehydrate it back in a single pass. 
If you don't need it (and you are clearly not) - then you might not need loadable.

In the same time - if you are already looking at prerendered-component - take a look at it's sibling package - [imported-component](https://github.com/theKashey/react-imported-component), it might be what you are looking for.

However - if you are still willing to proceed with this change - let's split it into smaller pieces to make it _reviewable_ and _mergeable_.

marcmrf:
> That is a very big PR which solves a problem which might be not a problem in the future.
> Upcoming React version gonna support suspense-based `partial hydration` out of the box, where you will be able to load `esm` modules with a simple `React.lazy`.
> 
> The reason to use loadable is to _track_ what you've used during SSR, to be able to properly rehydrate it back in a single pass.
> If you don't need it (and you are clearly not) - then you might not need loadable.
> 
> In the same time - if you are already looking at prerendered-component - take a look at it's sibling package - [imported-component](https://github.com/theKashey/react-imported-component), it might be what you are looking for.
> 
> However - if you are still willing to proceed with this change - let's split it into smaller pieces to make it _reviewable_ and _mergeable_.

Hi, the reason why i decided to propose this change is because of imported-component does not support `import(./${value})`. I'm aware react with concurrent mode is going to cover all these cases, but the release date is still a bit blurry. 

Our application currently uses both chunks and native imports. loadable-components cover some cases and react-imported-component some others. So, in order to be able to ssr properly everything, with the current solutions, we would need to use both of them, being difficult to maintain.

I'm open to invest time splitting the pr in several pieces if you think it make sense to be able to cover this case too from loadable-components

theKashey:
Give me some time to understand the changes here.

theKashey:
wait a little stalebot

keenwon:
I have the same question.
Is there a conclusion that the discussion here is?

PR #575, opened 2020-05-24
Add example for async-node target

7rulnik:
## Summary

Webpack 5 introduces new target `async-node`. It uses fs and vm to load chunks asynchronously.

More details you can find in webpack sources. Take a look on [`NodeTemplatePlugin`](https://github.com/webpack/webpack/blob/f06086c53b2277e421604c5cea6f32f5c5b6d117/lib/WebpackOptionsApply.js#L110) and [`StartupChunkDependenciesPlugin`](https://github.com/webpack/webpack/blob/f06086c53b2277e421604c5cea6f32f5c5b6d117/lib/WebpackOptionsApply.js#L119).

Seems that we should wait promise from [`ctor.requireAsync`](https://github.com/gregberge/loadable-components/blob/338bf555adc68986b12c8dd4e304875425119ca2/packages/component/src/createLoadable.js#L72), but for now we can't do it because it's inside of constructor. And even more: is it possible at all to wait for it without suspense on server? If it's so we just can go through `namedChunkGroups` from stats and  call `__webpack_require__.f.readFileVm` or `__webpack_require__.e`, but not sure that it will help

Note that this branch already contains fix from https://github.com/gregberge/loadable-components/pull/568

## Test plan

1. `cd examples/server-side-rendering-async-node`
2. `yarn`
3. `yarn build`
4. `yarn start`
5. Open page and see `TypeError: e[t] is not a function` in console


7rulnik:
As I can see with  `async-node` we also have the problem with mini-css-extract-plugin. It tries to load CSS in an async way, but it [works](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/b197757e26af717a302485293a2b95bc0eb6cf71/src/index.js#L336) only in a browser. @evilebottnawi is it a known issue?

evilebottnawi:
@7rulnik known issue

>  But as I understand we should not use it in server config. Right?

You can use it :smile:

theKashey:
Look like async node will not be able to work for now. 

As long as we need everything to be resolved before render, and solution before was bound to the ability to resolve all we need during the render - that's not gonna work anymore.
So the only way forward `node-async` is calling __all possible imports__ and awaiting for them.

That has two problems:
- is there any reason to use `node-async` and EVERYTHING would be resolved in any way?
- with `full dynamic import` it's not always known what you might import - that information is not exposed, however known for babel plugin.

So, question - what is the reason to use `node-async`?

7rulnik:
Not sure, but maybe module federations is the reason for it? As for now, we have the issue  https://github.com/module-federation/module-federation-examples/issues/48 so we need restart server to fetch new modules. Not sure that `node-async` with some enhancement will help there. Maybe @ScriptedAlchemy have some thoughts?

theKashey:
Not stale

theKashey:
I am adding webpack5 related examples and gonna build a few integation tests on top of them. 
Let's start with this one.

ScriptedAlchemy:
Omg yay!

ScriptedAlchemy:
Once you have this running again. I'll extend the MF api so we can pass chunks along with the remote for cross container lookups. We would need to be able to merge additional context into your chunk extractor, from the remotes build stats. Then I'll have the container expose get, init, and chunkMap.

PR #576, opened 2020-05-29
Sort preload assets so that styles are first

tomkelsey:
## Summary

When getting the preload assets, sort them so that the style assets are first.

I saw a PR over in next.js (https://github.com/vercel/next.js/pull/9486). Apparently preloading the JS first can slow down some browsers and delay the styles loading and therefore negatively impact performance/lighthouse scores.

I tested the approach on our own site and it led to quite a dramatic improvement in lighthouse score.

## Test plan

I tried to run the tests locally but it errored with something I didn't understand. I'm afraid I don't have the time at the moment to wrap my head around it but figured this PR may be useful to some people so would submit it anyway.

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -19,7 +19,10 @@ function extensionToScriptType(extension) {
 }
 
 function getAssets(chunks, getAsset) {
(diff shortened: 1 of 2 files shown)


theKashey:
Good idea. We were so concerned making js faster to load that made CSS load slower. And it blocks page rendering, so has to be to top priority.

theKashey:
However no merges on Friday.

PR #607, opened 2020-08-04
Fix: #592

### LINKED ISSUE #592: `chunkExtractor.getScriptElements` returns nested array in 5.13.1
## 🐛 Bug Report

Public API for `chunkExtractor.getScriptElements` is described as being an Array of React <script> Elements. As of version `5.13.1` it returns an array who's first element is also an array.

## To Reproduce

Unfortunately I cannot share my private repo. But the recent change to this function was the cause of this error. https://github.com/gregberge/loadable-components/blob/8cf392d124100763921226bdfcac10d350b64b34/packages/server/src/ChunkExtractor.js#L300 

Previously it returned a single element, but is now returning an array that isn't getting spread in this function 

https://github.com/gregberge/loadable-components/blob/8cf392d124100763921226bdfcac10d350b64b34/packages/server/src/ChunkExtractor.js#L384

## Expected behavior

`chunkExtractor.getScriptElements` returns an array of react script elements.

```bash

```

markwoodward23:
## Summary

Fixes #592 

## Test plan

Unit tests all pass + local linking.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,22 +40,28 @@ _Before_ submitting a pull request, please make sure the following is done…
     yarn --version
     ```
 
-4.  If you've added code that should be tested, add tests. You can use watch mode that continuously transforms changed files to make your life easier.
+4. Run `yarn build` to bootstrap packages.
+
+    ```sh
+    yarn build
+    ```
+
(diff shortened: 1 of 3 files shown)


markwoodward23:
How do we get this released @gregberge?

theKashey:
🤜 starting preparing the new release

PR #614, opened 2020-08-16
Update faq.mdx with link to React on Rails Pro

justin808:
## Summary

Let those considering loadable components know there is a Ruby on Rails solution in production.

## Test plan

n/a


theKashey:
👍

PR #619, opened 2020-08-22
Added onload callback option and fixed git-release

asaveliev:
## Summary
When loading the components, the component creator has no information regarding the state of the component.
Due to async nature of loading, the parent componentDidUpdate is called before lazy-loaded child is rendered. This makes state checks very complex - the only option is for the child component to emit some custom "Hey I rendered!" action/event.
This PR simply adds a callback to the options.

Note that after the component is loaded one should actually start using componentDidUpdate, as subsequent re-renders of a component are not passing via this onload callback. So basically you have to statefully track if a component has been rendered once. This state tracking may be implemented in loadable-components eventually.

## Test plan

I am using the updated library in my project, and this works fine. I added a unit test that checks this.
Since I am using my own fork, i had to make a package and fix the git-release.sh script - didn't seem like this was working.

### DIFF
--- a/packages/component/rollup.config.js
+++ b/packages/component/rollup.config.js
@@ -12,6 +12,7 @@ const name = 'loadable'
 const globals = {
   react: 'React',
   'hoist-non-react-statics': 'hoistNonReactStatics',
+  "react-is": "react-is",
 }
 
 const external = id => !id.startsWith('.') && !id.startsWith('/')
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -150,6 +150,11 @@ function createLoadable({
             onLoad(this.state.result, this.props)
           })
         }
+        if (options.onLoad){
+          setTimeout(() => {
+            options.onLoad()
+          })
+        }
       }
 
       loadSync() {
--- a/packages/component/src/loadable.test.js
+++ b/packages/component/src/loadable.test.js
@@ -76,6 +76,17 @@ describe('#loadable', () => {
     await wait(() => expect(container).toHaveTextContent('loaded'))
   })
 
+  it('should call onload callback option', async () => {
+    const load = createLoadFunction()
+    const onLoadCallback = jest.fn();
+    const Component = loadable(load, { onLoad: onLoadCallback })
+    const { container } = render(<Component />)
+    expect(container).toBeEmpty()
+    load.resolve({ default: () => 'loaded' })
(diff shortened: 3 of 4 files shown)


theKashey:
Got you. It would be easier to make this via `createLoadable` which already has `onLoad` props, but it is not exported.

In any case I would probably not recommend such approach, especially to gathering existing HTML.
What about another way:
```
const LoadableComponent = lazy(() => import('something'));

<Suspense fallback="loading">
 <CaptureHTML id="some-cache-id">
   <LoadableComponent/>
 </CaptureHTML>
</Suspense>

const CaptureHTML = ({id, children}) => {
  const divId = useUID();
  useEffect(() => {
    storeIt(id, document.getElementById(divId));
  }, []);
 
 return <div id={id}>{children}</div>
};
```

The idea is that `Suspense` would "unblocked" and `CaptureHTML`'s `useEffect` would be called only when __loadable become ready__.

PS: On SSR you can do `React.Suspense = React.Fragment`
PPS: CaptureHTML is not working on SSR. If you need fragment caching on SSR consider [react-preprendered-component](https://github.com/theKashey/react-prerendered-component)

ParallelUniv3rse:
@theKashey Correct me if I'm wrong but I feel like the proposed solution with Suspense wouldn't work. 
The Suspense gets "Unblocked" when the import promise gets resolved, not when the actual resolved component mounts. 

After the import gets resolved, the fallback component still gets mounted first, due to `loading: true` and `result: null` in the state. 
Only after the fallback gets mounted the loadAsync() fires, setting the correct state and rendering & mounting the `result` component. 
At that point it's too late and `<CaptureHTML>` has already executed it's hook. 

I'm fighting with this behavior myself currently, trying to animate a loadable component that was previously prefetched. 
Although it's been prefetched, it does not mount straight away and gets replaced after the first mount of fallback, thus making the react-transition-group animation impossible.

Would be really nice to have access to the onLoad for this use case as well.

theKashey:
@ParallelUniv3rse two things here:
- Its not  _The Suspense gets "Unblocked" when the import promise gets resolved__ - `Suspense` will block if something inside it __throws__. So the provided code works.
- `Suspense` in such form __is not working__ since 16.10. See https://github.com/theKashey/react-imported-component/issues/201#issuecomment-690152823
- long story short - it will render `failback`. To make it work you have to omit `failback` during hydration.

There is no in-house solution for this yet, except using React 16.9 or below (or experimental React, where Suspense does partial hydration by it's own)

Issue #636, opened 2020-09-29
Support Webpack 5


WaveString:
## 🐛 Bug Report

When I build with webpack 5.0.0-rc.0, I have an error `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Object`, because in `loadable-stats.json` field `assets` generated as  
```
 "assets": [{ "name": "assetname.js" }]
```
but `chunkExtractor` expects
```
"assets": ["assetname.js"]
```
## To Reproduce

1. Use webpack 5.0.0-rc.0
2. loadable-stats:
```
 "assets": [
        {
          "name": "assetname.js"
        }
  ]
```

## Expected behavior

Support new schema of assets

```bash
## System:
 - OS: macOS 10.15.6
 - CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
 - Memory: 1.45 GB / 16.00 GB
 - Shell: 5.7.1 - /bin/zsh
## Binaries:
 - Node: 14.9.0 - ~/.nvm/versions/node/v14.9.0/bin/node
 - Yarn: 1.22.4 - /usr/local/bin/yarn
 - npm: 6.14.8 - ~/.nvm/versions/node/v14.9.0/bin/npm
## npmPackages:
 - @loadable/babel-plugin: 5.13.2 => 5.13.2 
 - @loadable/component: 5.13.2 => 5.13.2 
 - @loadable/server: 5.13.2 => 5.13.2 
 - @loadable/webpack-plugin: 5.13.0 => 5.13.0 

```

jandvik:
Is there any traction on this? Its a pretty straight forward fix and necessary if anybody wants to use this with Webpack 5 Module Federation.

Easy solution with backward compatibility (might not be the prettiest)
```
createChunkAsset({ filename, chunk, type, linkType }) {
    let cleanFilename = typeof filename === 'object' && filename.name ? filename.name : filename
    return {
      cleanFilename,
      scriptType: extensionToScriptType(
        path
          .extname(cleanFilename)
          .split('?')[0]
          .toLowerCase(),
      ),
      chunk,
      url: this.resolvePublicUrl(cleanFilename),
      path: path.join(this.outputPath, cleanFilename),
      type,
      linkType,
    }
  }
```

I'd be happy to submit a PR for it, however I could not find any information on contributing.

bertho-zero:
This DeprecationWarning remains with the webpack plugin:

```
[...]

PR #638, opened 2020-10-12
feat: make packages webpack 5 compatible

StephaneRob:
## Summary

make packages compatible with Webpack 4 and Webpack 5

### DIFF
--- a/packages/component/.size-snapshot.json
+++ b/packages/component/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/loadable.cjs.js": {
-    "bundled": 14563,
-    "minified": 7172,
-    "gzipped": 2534
+    "bundled": 16387,
+    "minified": 7324,
+    "gzipped": 2574
   },
   "dist/loadable.esm.js": {
-    "bundled": 14184,
-    "minified": 6868,
-    "gzipped": 2467,
+    "bundled": 16008,
+    "minified": 7020,
+    "gzipped": 2506,
     "treeshaked": {
       "rollup": {
         "code": 276,
         "import_statements": 276
       },
       "webpack": {
-        "code": 5744
+        "code": 5896
       }
     }
   }
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -201,17 +201,20 @@ class ChunkExtractor {
   }
 
   createChunkAsset({ filename, chunk, type, linkType }) {
+    const resolvedFilename =
+      typeof filename === 'object' && filename.name ? filename.name : filename
+
     return {
-      filename,
+      filename: resolvedFilename,
       scriptType: extensionToScriptType(
         path
-          .extname(filename)
+          .extname(resolvedFilename)
           .split('?')[0]
           .toLowerCase(),
       ),
       chunk,
-      url: this.resolvePublicUrl(filename),
-      path: path.join(this.outputPath, filename),
+      url: this.resolvePublicUrl(resolvedFilename),
+      path: path.join(this.outputPath, resolvedFilename),
       type,
       linkType,
     }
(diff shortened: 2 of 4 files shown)


theKashey:
Fixes #636. PS @ScriptedAlchemy this is the fix you were looking for.

theKashey:
@StephaneRob - just it time. Thank you so much for this contribution.

ajayjaggi:
Hi.Even after these changes loadable is failing in SSR for federated modules while using webpack 5. I have also raised the issue.

ScriptedAlchemy:
Federated modules need to be handled a specific way. I don't think Babel can extract chunk mappings. @theKashey I'll update my ssr example then let's see what's needed for MF support.

ajayjaggi:
Thanks @ScriptedAlchemy . I saw your Repo https://github.com/module-federation/module-federation-examples/tree/master/server-side-render-only. It works when @loadable fixes are merged into this.But there is no example where components are imported using loadable function.It fails for me there.
It would be a great help if you could look into this.
Thanks.

https://github.com/ajayjaggi/MicroForntEnd-Basic-Structure - This is my repo where i have tried this.

ScriptedAlchemy:
Ahhh. Yeah that's because loadable probs tried to babelify the import into a hoc. But there's no chunkmapping so hydration issues can happen. 

You could try enabling top level await. But I think chunk flushing might still be a problem.

ScriptedAlchemy:
Same issue I ran into with next. I've been using lazy hydration so the component remains static till webpack runtime is booted, then I'd resolve the import client side and allow the component to hydrate

ajayjaggi:
https://github.com/ajayjaggi/MicroForntEnd-Basic-Structure - this is the PR for project.

The error is occurring only in server side rendering. For client side everything is working fine
In this file I have used loadable to load a file which is importing a federated module  - https://github.com/ajayjaggi/MicroForntEnd-Basic-Structure/blob/master/Edge/src/shared/routes.js .
 Also I have used top level await during SSR.

theKashey:
Meanwhile - can we move from merged PR to an open issue -> #638

ScriptedAlchemy:
If you have the remotes on the page immediately. Like hardcoded. All remote code should transport down in a single RTT. 

If you want to SSR the chunks a remote requires: 

1) loadable needs a new way to map names of remotes to federated chunk maps. 
2) I need to extend the MF api so remotes will export their chunk maps so loadable can extract them during render. 

Whenever you use loadable. Babel basically transforms the import into a object with a normal require statement and the module / chunk id. Then it's executed the HOC pushes these IDs into a scope that the server can access after the react render. 

The only Id it will have is ./someExposedModule - and we would need to go over the remotes stats and find what the browser side script is for that module. 

Same problem exists in next js. The loadable method straight crashes. 
No existing code split tool is capable of MF chunk coronations. I work around hydration issues with top level awaits. Which work on both client and server. 

You could also use partial hydration and simply hydrate markup when visible. 

I don't foresee this being hard to accomplish. Especially since I can extend MF apis at will. 

We could even change the getter on the remote so whenever a server gets code from the remote, it's container will report what request was asked for from the remote. It actually could be done without and need for Babel at all, or even loadable HOC. The only part we would need to adapt is ChunkExtractor to read from the second scope the remote containers are accessing directly. 

This mechanism would work exactly like react-universal does. Push right into a map, but we will be doing this inside the webpack runtime directly.

HugoGiraudel:
Out or curiosity—and sorry for interjecting in the middle of the conversation, is there a reason this update has not been published on npm already?

theKashey:
the absence of a webpack5 based automated test to make this release a little more secure

theKashey:
During the testing, I've actually found a real blocker for `loadable` usage. But it's not on the loadable side - so I am releasing a __beta__ version of loadable to let others test it as well.

Issue #652, opened 2020-10-24
Webpack plugin: doesn't ignore hot-update.js files from hot reload


xnerhu:
## 🐛 Bug Report

File loadable-stats.json is unnecessarily large, because it has hot reload chunks. 

## To Reproduce

Here is my webpack config

```js
const { resolve } = require('path');
const webpack = require('webpack');
const LoadablePlugin = require('@loadable/webpack-plugin');

module.exports = getConfig({
  output: {
    chunkFilename: '[name].chunk.js',
    hotUpdateMainFilename: 'updates/[hash].hot-update.json',
    hotUpdateChunkFilename: 'updates/[id].[hash].hot-update.js',
  },

  plugins: [new LoadablePlugin(), new webpack.HotModuleReplacementPlugin()],
});
```

And some of loadable-stats.json

```json
  "assetsByChunkName": {
    "main": [
      "main.js",
      "main.6993d44c5f54db65f33b.hot-update.js"
    ],
    "account": [
      "account.js"
    ],
    "history": [
      "history.js",
      "history.6993d44c5f54db65f33b.hot-update.js"
    ]
  },
```

## Details

```bash
## System:
 - OS: Windows 10 10.0.19041
 - CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor        

 - Memory: 8.38 GB / 15.95 GB
## Binaries:
 - Node: 12.16.3 - C:\Program Files\nodejs\node.EXE       
 - Yarn: 1.22.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD - npm: 6.14.5 - C:\Program Files\nodejs\npm.CMD
## npmPackages:
 - @loadable/babel-plugin: ^5.13.2 => 5.13.2
 - @loadable/component: ^5.14.1 => 5.14.1
 - @loadable/server: ^5.14.0 => 5.14.0
 - @loadable/webpack-plugin: ^5.14.0 => 5.14.0
 - webpack: 5.2.0
```

PR #655, opened 2020-11-02
feat: filter chunk assets

cahnory:
## Summary

In development, when doing SSR with HMR you don't want the hmr related files to be linked by the served html. The `filter` option in addition to the new `info` property added to chunk assets make it possible.

I favored this approach rather than an _"excludeHmrFiles"_ type option as it could serve other purposes and remain relevant even if hmr is abandoned in the future.

## Test plan

Using webpack with hmr enabled, instanciate your `ChunkExtractor` like :

```js
import { ChunkExtractor } from '@loadable/server'

const extractor = new ChunkExtractor({
  statsFile: '../dist/loadable-stats.json',
  filter: ({ info }) => !info.hotModuleReplacement,
});
```

All files relative to hmr (suffixed with _"-wps-hmr"_ by default) should be excluded from the results of `getScriptTags`, `getScriptElements`,…

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -202,7 +202,7 @@ class ChunkExtractor {
     namespace = '',
     outputPath,
     publicPath,
-    filter,
+    filterChunk,
     inputFileSystem = fs,
   } = {}) {
     this.namespace = namespace
@@ -211,7 +211,7 @@ class ChunkExtractor {
     this.outputPath = outputPath || this.stats.outputPath
     this.statsFile = statsFile
     this.entrypoints = Array.isArray(entrypoints) ? entrypoints : [entrypoints]
-    this.filter = filter || Boolean
+    this.filterChunk = filterChunk || Boolean
     this.chunks = []
     this.inputFileSystem = inputFileSystem
     this.assetsByName = this.stats.assets.reduce((acc, asset) => {
@@ -237,7 +237,7 @@ class ChunkExtractor {
   }
 
   isExctractableChunkAsset(chunk) {
-    return isValidChunkAsset(chunk) && this.filter(chunk)
+    return isValidChunkAsset(chunk) && this.filterChunk(chunk)
   }
 
   createChunkAsset({ filename, chunk, type, linkType }) {
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -202,6 +202,7 @@ class ChunkExtractor {
     namespace = '',
     outputPath,
     publicPath,
+    filter,
     inputFileSystem = fs,
   } = {}) {
     this.namespace = namespace
@@ -210,8 +211,13 @@ class ChunkExtractor {
     this.outputPath = outputPath || this.stats.outputPath
     this.statsFile = statsFile
     this.entrypoints = Array.isArray(entrypoints) ? entrypoints : [entrypoints]
+    this.filter = filter || Boolean
     this.chunks = []
     this.inputFileSystem = inputFileSystem
+    this.assetsByName = this.stats.assets.reduce((acc, asset) => {
+      acc[asset.name] = asset
+      return acc
+    }, {})
   }
 
   resolvePublicUrl(filename) {
@@ -230,9 +236,15 @@ class ChunkExtractor {
     return chunkInfo
   }
 
+  isExctractableChunkAsset(chunk) {
+    return isValidChunkAsset(chunk) && this.filter(chunk)
+  }
+
   createChunkAsset({ filename, chunk, type, linkType }) {
     const resolvedFilename =
       typeof filename === 'object' && filename.name ? filename.name : filename
+
+    const asset = this.assetsByName[resolvedFilename]
 
     return {
       filename: resolvedFilename,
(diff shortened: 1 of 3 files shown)


theKashey:
Fixes #652 ?

packages/server/src/ChunkExtractor.js
```diff
@@ -187,8 +188,13 @@ class ChunkExtractor {
     this.outputPath = outputPath || this.stats.outputPath
     this.statsFile = statsFile
     this.entrypoints = Array.isArray(entrypoints) ? entrypoints : [entrypoints]
+    this.filter = filter || Boolean
```
theKashey, line 191: Let's use a less generic name on `this` scope. For example `filterChunk`

cahnory:
> Fixes #652 ?

I don't think so.

I didn't want to mess with _loadable-stats.json_ as my intent was only to prevent serving hmr chunks but the stats could potentially serve other purposes. 
So this PR allows `ChunkExtractor` to apply a filter on the assets it returns via its getter methods while #652 would requires the webpack plugin to apply a filter on the chunks it add to the _loadable-stats.json_ file.

packages/server/src/ChunkExtractor.js
cahnory: You're right, I made the change.

cahnory: I also rename the option as it makes its purpose more explicit.

cahnory:
Don't know why the ci failed on the last commit.

cahnory:
@theKashey if you prefer that the filter occurs in the webpack plugin (exclude chunks from stats file) let me know.

packages/server/src/ChunkExtractor.test.js
```diff
@@ -299,6 +307,55 @@ describe('ChunkExtrator', () => {
...
+            async={true}
+            data-chunk="letters-A"
+            nonce="testnonce"
+            src="/dist/node/7415d90-letters-A-wps-hmr.js"
+          />,
+        ]
+      `)
+    })
+
+    it.only('should use filter from options', () => {
```
theKashey, line 320: `.only` should be removed 😅

theKashey (changes requested):

packages/server/src/ChunkExtractor.test.js
cahnory: Woopsy ^^

cahnory:
@theKashey I'm having a problem with _stats.json_.

 When generated, it does not contain hmr chunks (obviously) and the `assets.*.info` property is always empty (but maybe the `hotModuleReplacement` property is only set when `true`).
I can manipulate stats object in my test to match my needs but I think it's not the way I'm supposed to do this.

theKashey:
Yeah, probably you should take the original stats (as it's now generated from a real project) and add some "extra" fields in runtime to test, or filter some "normal" chunks, as it does not have test `hot` case, only the filtering

cahnory:
> or filter some "normal" chunks, as it does not have test `hot` case, only the filtering

I had my initial usecase so much on my mind that I didn't think about it!

theKashey:
👀 .....

cahnory:
😅 I don't know what's wrong. I run the commands `ci`, `test` and even `build` locally without any problem. It seems to me that it has something to do with a deployment step that I don't master. In netlify logs I just see :
> failed Building production JavaScript and CSS bundles

Any idea ?

theKashey:
@cahnory - it's just dead a little. Strangely there is no travis here, and these tests are the only really important thing.

eliseumds:
Anyway I can help here?

theKashey:
The real problem here is a "new" undocumented feature, which implementation is still on the user.
If HMR temporal files should not be a part of webpack builds - then they should not. Full stop and no extra conditions.

cahnory:
> The real problem here is a "new" undocumented feature, which implementation is still on the user.
> If HMR temporal files should not be a part of webpack builds - then they should not. Full stop and no extra conditions.

This is not what I understood from our previous exchanges. In this case, if I understand correctly, it seems unsolvable and maybe we should close this PR.

theKashey:
> This is not what I understood from our previous exchanges.

Decisions change in time. That's ok.

>  if I understand correctly, it seems unsolvable

The internal changes are legit and I would like to have them. I am just thinking that we should, at least for now, put `filter: ({ info }) => !info.hotModuleReplacement,` __inside__ loadable logic, keep external API intact as and I am not sure is there any other use case for `filter`. If you have one - feel free to change my mind.

The _lazy maintainer rule_ is simple - if you can __not expose__ something - do not, or you will have to maintain it forever (and you are lazy by definition).

cahnory:
I understand, no problem. Just that by doing this way this PR should be closed in favor of a new one which maybe just add a cookbook recipe to the doc.

The way I do it right now is just by filtering `extractor.getXElements` results with this function:
```js
const filterHmr = (list) =>
  list.filter(
    (element) =>
      !element.props.src?.match(/wps-hmr\.[^.\\/]+$/) &&
      !element.props.href?.match(/wps-hmr\.[^.\\/]+$/),
  );
```

I just wasn't satisfied with having to check the file names rather than their stats (and extra loops but they are negligible).

theKashey:
That's why it's still good to have the majority of this PR built-in into loadable with the correct behavior working out of the box

cahnory:
> That's why it's still good to have the majority of this PR built-in into loadable with the correct behavior working out of the box

Oh I see, I completely misunderstood what you were saying. Sorry.
 I'll try to find a moment to make the changes.

Issue #248, opened 2019-02-20
Cannot find chunk in stats


nandosangenetto:
## 🐛 Bug Report

I'm having a problem similar to the #216 problem. I run webpack and it gets all chunks, but the server cannot find the chunk in the stats json.

My `babel.config.js` is:
```javascript
function isWebTarget(caller) {
  return Boolean(caller && caller.target === 'web');
}

function isWebpack(caller) {
  return Boolean(caller && caller.name === 'babel-loader');
}

module.exports = api => {
  const web = api.caller(isWebTarget);
  const webpack = api.caller(isWebpack);

  return {
    presets: [
      '@babel/preset-react',
      [
        '@babel/preset-env',
        {
          useBuiltIns: web ? 'entry' : undefined,
          targets: !web ? { node: 'current' } : undefined,
          modules: webpack ? false : 'commonjs',
        },
      ],
    ],
    plugins: [
      '@babel/transform-runtime',
      '@babel/transform-async-to-generator',
      '@babel/proposal-class-properties',
      'babel-plugin-dynamic-import-webpack',
      '@loadable/babel-plugin',
      'styled-components',
    ].filter(Boolean),
  };
};
```

I've tried to reorder the plugins and adding `@babel/plugin-syntax-dynamic-import` which doesn't change anything. If I remove the `babel-plugin-dynamic-import-webpack` my app doesn't build.

And following the config above I was able to successfully build the project but when someone try to access the server, I get this error:

```
>>> Req.path:  /
(node:23605) UnhandledPromiseRejectionWarning: Invariant Violation: loadable: cannot find components-HomePage in stats
    at invariant (/Users/nando.sangenetto/Sites/reverb-web/node_modules/@loadable/component/dist/loadable.cjs.js:13:15)
    at ChunkExtractor.getChunkGroup (/Users/nando.sangenetto/Sites/reverb-web/node_modules/@loadable/server/lib/ChunkExtractor.js:175:36)
    at one (/Users/nando.sangenetto/Sites/reverb-web/node_modules/@loadable/server/lib/ChunkExtractor.js:246:31)
    at arrayMap (/Users/nando.sangenetto/Sites/reverb-web/node_modules/lodash/lodash.js:639:23)
    at map (/Users/nando.sangenetto/Sites/reverb-web/node_modules/lodash/lodash.js:9556:14)
    at Function.flatMap (/Users/nando.sangenetto/Sites/reverb-web/node_modules/lodash/lodash.js:9259:26)
[...]

PR #664, opened 2020-11-24
Add section about compatibility with other plugins

lostpebble:
Specifically the `babel-plugin-dynamic-import-webpack`, which some people might still be using.

## Summary

Some people have run into issues, as shown in #248

### DIFF
--- a/website/src/pages/docs/babel-plugin.mdx
+++ b/website/src/pages/docs/babel-plugin.mdx
@@ -117,3 +117,7 @@ const OtherComponent = loadable(loadOther)
 ```
 
 The `loadOther` function will be transformed into an object, this way you can manipulate function and it is still portable!
+
+## Compatibility Considerations
+
+If you still happen to be using `babel-plugin-dynamic-import-webpack` you should remove it because it is not compatible with this plugin. Newer versions of Webpack now understand dynamic imports so you don't need this older plugin in any case.


theKashey:
Hey. Thank you for the PR, but I don't think it will solve the problem.

It's a little strange that so many people are using it, but adding some comments in the docs would not help them understand their problem - however, adding some comments in the github issue might be a better option.

Can you share your own experience and the way you found #248? Would documentation update change it?

What can be nice - is to be able to automatically to detect the problem with the environment and inform the user about it. Like
```
 Invariant Violation: loadable: cannot find components-HomePage in stats 👉, for example because of this or that
```

Loadable itself is not able to detect `imports` modifications, as it works "before" any plugin to actually allow other plugins to transform `import` to something else, which is still needed for `nodejs` for example.

So, what about updating the error message instead?
Keep in mind - it shall not be bound to `babel-plugin-dynamic-import-webpack` as it just one of many possible ways to break imports/splitting. There is an easier way to break everything using a wrong webpack configuration.

lostpebble:
Hi @theKashey ,

Agreed, giving an error message that points you in the right direction would always be better than an obscure note in the docs. Just figured that it can't hurt, and would have saved me time as I was going through all the docs trying to figure out if I had a compatibility issue but there wasn't any mention of anything.

I think if there was just more information that there could be possible clashes with other plugins, that could help already and people would search the issues for that. Or have the error message point to a section in the docs about compatibility (perhaps in the SSR section)? Which allows you to expand on what might be happening.

> Can you share your own experience and the way you found #248? Would documentation update change it?

I searched the issues for the error message (well, the parts of it that are generic) - "cannot find in stats" which popped up a bunch of issues- one of which had the actual information I needed (#248).

theKashey:
> I searched the issues for the error message ... had the actual information I needed (#248).

Sounds like this is how everyone else will solve their problem. It can be really hard to go through all the loadable docs and not miss that small information about compatibility.

lostpebble:
> Sounds like this is how everyone else will solve their problem. It can be really hard to go through all the loadable docs and not miss that small information about compatibility.

Understood, though a little more direction would have gone a long way. I almost didn't find the issue at hand, because there were many which have the same error message structure.

> I think if there was just more information that there could be possible clashes with other plugins, that could help already and people would search the issues for that. Or have the error message point to a section in the docs about compatibility (perhaps in the SSR section)? 

Just having a more general compatibility section or error message saying that it may clash with other dynamic import plugins would have gone a long way.

PR #676, opened 2020-12-17
fix: update hooks for webpack 5

bertho-zero:
Related to https://github.com/gregberge/loadable-components/issues/636

PR #677, opened 2020-12-21
fix: DeprecationWarning Compilation.assets

duterte:
## Summary

This is to fix Webpack 5 deprecation warning :

```terminal
(node:4312) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
        Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
        Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
```

This warning was trace to [line 32](https://github.com/gregberge/loadable-components/blob/89cf124f9079395f2d77038033e472b22c04deb6/packages/webpack-plugin/src/index.js#L32) of packages/webpack-plugin/src/index.js. This warning will fired when plugin emit `loadable-stats.json`

## Test plan

This version of code is compatible with Webpack version 4 and 5. I've already used in my own  project. Original version uses compilation.assets to emit `loadable-stats.json` . This version will produce similar output by not using compilation hook. This should work even `compilation.getStats().toJson()` produces huge bytes of strings.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -29,14 +29,13 @@ class LoadablePlugin {
     const result = JSON.stringify(stats, null, 2)
 
     if (this.opts.outputAsset) {
-      hookCompiler.assets[this.opts.filename] = {
-        source() {
-          return result
-        },
-        size() {
-          return result.length
-        },
-      }
+      const outputPath = nodePath.resolve(
+        this.compiler.options.output.path,
+        this.opts.filename,
+      )
+      const file = fs.createWriteStream(outputPath)
+      file.write(result)
+      file.end()
     }
 
     if (this.opts.writeToDisk) {
@@ -50,7 +49,7 @@ class LoadablePlugin {
    * Write Assets Manifest file
    * @method writeAssetsFile
    */
-  writeAssetsFile = manifest => {
+  writeAssetsFile = (manifest) => {
     const outputFolder =
       this.opts.writeToDisk.filename || this.compiler.options.output.path
 


packages/webpack-plugin/src/index.js
```diff
@@ -29,14 +29,13 @@ class LoadablePlugin {
...
-        size() {
-          return result.length
-        },
-      }
+      const outputPath = nodePath.resolve(
+        this.compiler.options.output.path,
+        this.opts.filename,
+      )
+      const file = fs.createWriteStream(outputPath)
+      file.write(result)
```
theKashey, line 37: this is definitely not a correct way to handle this moment. You should not `write` file in order to handle deprecation warning, especially if that functionality already exists via writeAssetsFile.

Ask yourself why there were two ways of doing this?

Let's proceed with, which addresses the same problem without removing breaking the contract between loadable and webpack. 
https://github.com/gregberge/loadable-components/pull/676/files#diff-ca9a5ec2c1a865ca7d717cfaed22c11c91459fefce687dccc92ca910c1409c5eL32

theKashey (changes requested):
Let's focus on a solution from #676

packages/webpack-plugin/src/index.js
duterte: https://github.com/gregberge/loadable-components/pull/676/files#diff-ca9a5ec2c1a865ca7d717cfaed22c11c91459fefce687dccc92ca910c1409c5eL32 This version works fine for me now, it bundle files without warning.

theKashey:
Closing as #676 got merged

PR #686, opened 2021-01-12
fix: remove main asset path query

jas0ncn:
## Summary

`requireEntrypoint` failed when main asset's path has query.

image


theKashey:
👍, as it was already done in a few places.

PR #689, opened 2021-01-13
fix: ignore single css chunk (close #688)

### LINKED ISSUE #688: LoadableReady not be called if required chunks have a chunk only contains css.
## 🐛 Bug Report

If webpack generates a chunk only contains css file, no js, loadable ready check will fail, and callback function not be called.

https://github.com/gregberge/loadable-components/blob/dde2252136a2d7f3ff635f85deff429948130883/packages/component/src/loadableReady.js#L56-L67

## To Reproduce

create a new react webpack project, just simple import css in components, add `mini-css-extract-plugin` to webpack config, insure pack a chunk only contains css file.

## Expected behavior

Filter out the chunk only contains css file when generating `RequiredChunksScriptTag`:

https://github.com/gregberge/loadable-components/blob/dde2252136a2d7f3ff635f85deff429948130883/packages/server/src/ChunkExtractor.js#L282-L284

Issues without a reproduction link are likely to stall.

## Run `npx envinfo --system --binaries --npmPackages @loadable/component,@loadable/server,@loadable/webpack-plugin,@loadable/babel-plugin --markdown --clipboard`

Paste the results here:

```bash
## System:
 - OS: macOS 11.0.1
 - CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
 - Memory: 1.62 GB / 16.00 GB
 - Shell: 5.8 - /bin/zsh
## Binaries:
 - Node: 14.15.2 - /var/folders/9t/xn8r9ffn31z6vnk1vmfdzcd00000gn/T/fnm_multishell_25701_1610416859901/bin/node
 - Yarn: 1.22.10 - /usr/local/bin/yarn
 - npm: 6.14.9 - /var/folders/9t/xn8r9ffn31z6vnk1vmfdzcd00000gn/T/fnm_multishell_25701_1610416859901/bin/npm
## npmPackages:
 - @loadable/babel-plugin: ^5.13.2 => 5.13.2 
 - @loadable/component: ^5.14.1 => 5.14.1 
 - @loadable/server: ^5.14.0 => 5.14.0 
 - @loadable/webpack-plugin: ^5.14.0 => 5.14.0
```

jas0ncn:
## Summary

Ignore single css chunk when `RequiredChunks` added into HTML output during SSR. 

Issue: https://github.com/gregberge/loadable-components/issues/688

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -171,6 +171,11 @@ function isValidChunkAsset(chunkAsset) {
   return chunkAsset.scriptType && !HOT_UPDATE_REGEXP.test(chunkAsset.filename)
 }
 
+const JS_FILE = /\.js$/
+function checkIfChunkIncludesJs(chunkInfo) {
+  return chunkInfo.files.some(file => JS_FILE.test(file.split('?')[0]))
+}
+
 class ChunkExtractor {
   constructor({
     statsFile,
@@ -199,6 +204,12 @@ class ChunkExtractor {
     const chunkGroup = this.stats.namedChunkGroups[chunk]
     invariant(chunkGroup, `cannot find ${chunk} in stats`)
     return chunkGroup
+  }
+
+  getChunkInfo(chunkId) {
+    const chunkInfo = this.stats.chunks.find(chunk => chunk.id === chunkId)
+    invariant(chunkInfo, `cannot find chunk (chunkId: ${chunkId}) in stats`)
+    return chunkInfo
   }
 
   createChunkAsset({ filename, chunk, type, linkType }) {
@@ -269,7 +280,17 @@ class ChunkExtractor {
   getChunkDependencies(chunks) {
     const one = chunk => {
       const chunkGroup = this.getChunkGroup(chunk)
-      return chunkGroup.chunks
+
+      // ignore chunk that only contains css files.
+      return chunkGroup.chunks.filter(chunkId => {
+        const chunkInfo = this.getChunkInfo(chunkId)
+
+        if (!chunkInfo) {
+          return false
+        }
+
+        return checkIfChunkIncludesJs(chunkInfo)
+      })
     }
 
     if (Array.isArray(chunks)) {
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -20,12 +20,19 @@ class LoadablePlugin {
       hash: true,
       publicPath: true,
       assets: true,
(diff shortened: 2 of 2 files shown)


theKashey:
There is just single one issue with test preventing merging this PR to a master branch 😭

packages/server/src/ChunkExtractor.js
```diff
@@ -171,6 +171,11 @@ function isValidChunkAsset(chunkAsset) {
   return chunkAsset.scriptType && !HOT_UPDATE_REGEXP.test(chunkAsset.filename)
 }
 
+const CSS_FILE = /\.css$/
```
cahnory, line 174: Rather than testing if all files are CSS, shouldn't it be tested if at least one file is JavaScript?

jas0ncn:
> There is just single one issue with test preventing merging this PR to a master branch 😭

Should I generate new `__fixtures__` folder to test this case, or just modify `stats.json` directly? This case needs webpack5 and I think maybe shouldn't update the server-side-rendering example to webpack5.

packages/server/src/ChunkExtractor.js
jas0ncn: Right, I will modify Regexp and test js.

cahnory:
> > There is just single one issue with test preventing merging this PR to a master branch 😭
> 
> Should I generate new `__fixtures__` folder to test this case, or just modify `stats.json` directly? This case needs webpack5 and I think maybe I shouldn't update webpack to v5 of the server-side-rendering example.

Not sure if it's the right answer but in my (now closed 😢) I chose to modify the fixtures:
https://github.com/gregberge/loadable-components/pull/655/files#diff-7979dacf4945ebda3d5b355e934617b7bf14dbe6ec1996030bce2d413503349a

I think if your code is working but not passing the tests, it's better to edit tests if possible (tests adapt to code, not the other way).

theKashey:
merging to continue development from my side...

theKashey:
> Should I generate new __fixtures__ folder to test this case, or just modify stats.json directly? 

fixtures should be always generated from some source code, but look like that moment got lost somewhere.
After regenerating status from the expected source(examples/server-side-rendering) I have 18 broken tests....

theKashey:
While there is no test yet for this use case, it will be very good to double check that current version in master works and/or modify https://github.com/gregberge/loadable-components/tree/master/examples/server-side-rendering to produce the _extended_ stats.
👉there is yet undocumented script `scripts/prepare` to do perform this operation.

Issue #693, opened 2021-01-25
Can't launch SSR example


astnt:
## 🐛 Bug Report

I'm trying to launch SSR example.

## To Reproduce

Follow https://github.com/gregberge/loadable-components/blob/master/examples/server-side-rendering/README.md instructions.

## Expected behavior

Example will start to work.

## Link to repl or repo (highly encouraged)

https://github.com/gregberge/loadable-components/tree/master/examples/server-side-rendering

Exception:

```
[nodemon] watching extensions: js,mjs,json
[nodemon] starting `babel-node src/server/main.js`
/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@babel/core/lib/config/files/plugins.js:175
    throw e;
    ^

Error: Cannot find module '/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@loadable/babel-plugin/lib/index.js'. Please verify that the package.json has a valid "main" entry
    at tryPackage (internal/modules/cjs/loader.js:318:19)
    at Function.Module._findPath (internal/modules/cjs/loader.js:700:18)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:963:27)
    at resolve (internal/modules/cjs/helpers.js:78:19)
    at resolveStandardizedName (/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@babel/core/lib/config/files/plugins.js:100:7)
    at resolvePlugin (/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@babel/core/lib/config/files/plugins.js:44:10)
    at loadPlugin (/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@babel/core/lib/config/files/plugins.js:52:20)
    at createDescriptor (/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@babel/core/lib/config/config-descriptors.js:154:9)
    at /Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@babel/core/lib/config/config-descriptors.js:109:50
    at Array.map (<anonymous>) {
  code: 'MODULE_NOT_FOUND',
  path: '/Users/user/WebstormProjects/loadable-components/examples/server-side-rendering/node_modules/@loadable/babel-plugin/package.json',
  requestPath: '@loadable/babel-plugin'
}
[...]

PR #695, opened 2021-01-26
Include build library step in SSR Example README

lmreicks:
The SSR example doesn't work without building the library, add that as a step for running the example.

## Summary

The SSR example doesn't work without building the library, add that as a step for running the example. Per issue: https://github.com/gregberge/loadable-components/issues/693

## Test plan

Not needed


theKashey:
👍

PR #708, opened 2021-02-28
[perf] Set all:false in stats.toJson call

liuyenwei:
## Summary
I noticed while upgrading a codebase from webpack v4 -> v5 that there seemed to be a significant increase in incremental build times. 

From a local profile, the time seemed to be spent on a `stats.toJson` call originating from the loadable webpack plugin
image

Updating the `toJson` call so that it specifies [all: false](https://webpack.js.org/configuration/stats/#statsall) improves the time spent on this call from 40seconds -> 55ms (for the specific codebase i tested on)
image

I checked this change against webpack 4 on our codebase and it also reduced the time spent from ~9s -> 500ms
image
image

## Test plan
I tested this change locally against both a webpack 4 build and webpack 5 build and loadable worked as expected in both cases.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -23,13 +23,11 @@ class LoadablePlugin {
       assets: true,
       chunks: true,
       chunkGroups: true,
-      errorDetails: false,
+      chunkGroupChildren: true,
       hash: true,
-      modules: false,
+      ids: true,
       outputPath: true,
       publicPath: true,
-      source: false,
-      timings: false,
     })
 
     stats.generator = 'loadable-components'
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -19,13 +19,16 @@ class LoadablePlugin {
 
   handleEmit = compilation => {
     const stats = compilation.getStats().toJson({
-      hash: true,
-      publicPath: true,
+      all: false,
       assets: true,
       chunks: true,
-      modules: false,
-      source: false,
+      chunkGroups: true,
       errorDetails: false,
+      hash: true,
+      modules: false,
+      outputPath: true,
+      publicPath: true,
+      source: false,
       timings: false,
     })
 


packages/webpack-plugin/src/index.js
```diff
@@ -19,13 +19,16 @@ class LoadablePlugin {
 
   handleEmit = compilation => {
     const stats = compilation.getStats().toJson({
-      hash: true,
-      publicPath: true,
+      all: false,
       assets: true,
       chunks: true,
+      chunkGroups: true,
+      errorDetails: false,
```
jdb8, line 26: Is there a reason to keep these `false` options provided? Or does `all: false` set that as a default for anything not explicitly provided?

bishwei: looking at https://github.com/webpack/webpack/blob/afc9b2fcf9bef0831640b3ebb02b73068ba18e17/lib/stats/DefaultStatsPresetPlugin.js#L132 and https://webpack.js.org/configuration/stats/#statsall, it doesn't seem so since `all: false` will basically default anything not set to `false`.

i opted for leaving these as-is since i was trying to make a minimal change here but happy to remove the unnecessary false options here

jdb8: IMO it'd be clearer to just totally omit anything that isn't `true`, since the `all: false` covers all that. But I'll defer to the repo maintainers, I'm just taking a look since this PR will solve my issue too (#709) 😄

packages/webpack-plugin/src/index.js
```diff
@@ -19,14 +19,15 @@ class LoadablePlugin {
...
-      hash: true,
-      publicPath: true,
+      all: false,
       assets: true,
       chunks: true,
-      modules: false,
-      source: false,
-      errorDetails: false,
-      timings: false,
+      chunkGroups: true,
```
bishwei, line 25: Updated to:
- remove the falsey options (not needed now that we're specifying all: false)
- added `chunkGroupChildren: true` and `ids: true`.  These are necessary for webpack 5 otherwise https://github.com/gregberge/loadable-components/blob/cba6cab47536e23ded8bac34ee24417e6452344e/packages/server/src/ChunkExtractor.js#L273 and https://github.com/gregberge/loadable-components/blob/main/packages/server/src/ChunkExtractor.js#L298 will fail

jdb8:
Looking good to me, but I'm not a maintainer - would love to see this merged to close #709 :)

bishwei:
@theKashey hey just wanted to check in and see if there was anything else I can/should do to help get this merged.  Thanks!

theKashey:
Just waiting for the Easter break to go through all changes

PR #730, opened 2021-03-30
Prevent memory leaks in staging by adding env var

OSamkova:
## Summary

clearModuleCache is causing memory leaks in staging. Added optional env var to prevent it.

## Test plan

Screen Shot 2021-03-30 at 5 35 05 PM

Screen Shot 2021-03-30 at 5 35 37 PM

### DIFF
--- a/packages/server/src/util.js
+++ b/packages/server/src/util.js
@@ -18,7 +18,7 @@ export const clearModuleCache = moduleName => {
 }
 
 export const smartRequire = modulePath => {
-  if (process.env.NODE_ENV !== 'production' && process.env.LOADABLE_KEEP_CACHE != 1) {
+  if (process.env.NODE_ENV !== 'production' && process.env.LOADABLE_KEEP_CACHE !== '1') {
     clearModuleCache(modulePath)
   }
 
--- a/packages/server/src/util.js
+++ b/packages/server/src/util.js
@@ -18,7 +18,7 @@ export const clearModuleCache = moduleName => {
 }
 
 export const smartRequire = modulePath => {
-  if (process.env.NODE_ENV !== 'production' && process.env.LOADABLE_KEEP_CACHE !== '1') {
+  if (process.env.NODE_ENV !== 'production' || process.env.LOADABLE_KEEP_CACHE !== '1') {
     clearModuleCache(modulePath)
   }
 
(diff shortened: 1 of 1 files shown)


theKashey:
I would ask you to open the issue first. 
It is absolutely not clear what you gonna fix, why it is a problem and why this change is a solution.

`smartRequire` is expected to clear module cache in dev mode to support hot module replacement, which never happens in the "non-dev" mode.
Adding extra constant(`LOADABLE_KEEP_CACHE)` which has to be enabled in order to clean cache will break the internal "contract" for all existing users and is a breaking change.

There is no way this PR can be merged as is.

theKashey (changes requested):
More information is needed

OSamkova:
Made a change and opened an issue as well.

theKashey:
Then #733 should have been explicitly mentioned in this PR

packages/server/src/util.js
```diff
@@ -18,7 +18,7 @@ export const clearModuleCache = moduleName => {
 }
 
 export const smartRequire = modulePath => {
-  if (process.env.NODE_ENV !== 'production') {
+  if (process.env.NODE_ENV !== 'production' || process.env.LOADABLE_KEEP_CACHE !== '1') {
```
theKashey, line 21: According to the ticket description, this is an __incorrect__ way to fix it, as `process.env.NODE_ENV !== 'production'` is still true, and the only change here - you are adding an ability to cleanup module cache in production is as well.

Pretty sure you want something else

theKashey, line 21: What about enabling clean cache only when HRM is enabled?
```suggestion
  if (process.env.NODE_ENV !== 'production' && module.hot) {
```

theKashey:
While bounding logic to `module.hot` seems to be right and as can move forward with it - there is still no information about the leak itself and why this is a fix.

OSamkova:
Adding module.hot did work, i checked it today. What kind of information should i provide about the memory leak? Would screenshots work? Without this our staging environment is crashing every 20 minutes or so.

theKashey:
Sounds like there is a memory leak in clearModuleCache, probably related to removing self from the children, creating an orphan -> https://github.com/gregberge/loadable-components/blob/4fb7d6dba5d97d22fade7f8a794b50d6506645af/packages/server/src/util.js#L12

OSamkova:
so should we move forward with module.hot logic fix?

theKashey:
Yes. That sounds as a must-have in any case

PR #735, opened 2021-04-02
feat: webpack plugin doesn't need all chunk info (150x faster)

eliseumds:
I'm currently working on a pretty massive project with hundreds of split points, thousands of modules (plus node_modules), CSS, hot reloading, all the goodies... but *LoadablePlugin* was taking a very significant amount of time to run (about 1.2s). With these changes it went down to ~7ms.


theKashey:
A good addition to #708

cizz3007:
wow

theKashey (approved):

PR #737, opened 2021-04-05
fix docs: import() syntax a part of language

SiarheiLunski:
## Summary

`import()` syntax is officially a part of ES2020 so this note can be refined.

https://github.com/tc39/proposal-dynamic-import

## Test plan

n/a


theKashey:
😎

PR #749, opened 2021-04-22
added option to exclude chunk from extractor

pajaron-mms:
## Summary

Motivation for this PR is to be able to completely exclude the chunk from loading if in case we would like to do performance optimizations.

For example:
- we would like to only render a component statically and therefore the additional JS is not needed.
- we would like to defer the loading of the chunk to go in line with our lazy-render so that it is never loaded initially if it isn't needed.

Please let me know if there could be a better naming, or what I need to do next to get this in.

See this working animation for the 2nd point:

## Test plan
ezgif com-video-to-gif-4

```
// loadable.js
if (!options.excludeChunkFromExtractor) {
  props.__chunkExtractor.addChunk(ctor.chunkName(props));
}

// component
const LazyFooter = loadable(() => import('./CmsFooter/CmsFooter'), {
  excludeChunkFromExtractor: true,
})
```

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -129,7 +129,9 @@ function createLoadable({
           // So we can require now the module synchronously
           this.loadSync()
 
-          props.__chunkExtractor.addChunk(ctor.chunkName(props))
+          if (!options.excludeChunkFromExtractor) {
+            props.__chunkExtractor.addChunk(ctor.chunkName(props))
+          }
           return
         }
 


theKashey:
Got you. A valid use case, using something like this as well for lazily hydrated stuff.

However, the naming is not that great:
- there are `ssr:false` prop, to "not render" something on the server side
- and there could be like `client:false`, to "not render" something on the client side
- 👉 and the mixed case, when _render on server_, but don't tell the client.

Can be done in two ways:
- what about `track:false` flag? `excludeChunkFromExtractor` is too specific, but what you want to do is to `not track`
- wrap some locations with another `ChunkExtractorManager`, throwing away collected information. In this case it will be a little safer if you have "nested" loadable - they all will be non tracked.

Thoughts?

pajaron-mms:
> Got you. A valid use case, using something like this as well for lazily hydrated stuff.
> 
> However, the naming is not that great:
> 
> * there are `ssr:false` prop, to "not render" something on the server side
> * and there could be like `client:false`, to "not render" something on the client side
> * 👉 and the mixed case, when _render on server_, but don't tell the client.
> 
> Can be done in two ways:
> 
> * what about `track:false` flag? `excludeChunkFromExtractor` is too specific, but what you want to do is to `not track`
> * wrap some locations with another `ChunkExtractorManager`, throwing away collected information. In this case it will be a little safer if you have "nested" loadable - they all will be non tracked.
> 
> Thoughts?

Thanks for the feedback! I will try and put a new `ChunkExtractorManager` in our `LazyRender` component and will see how it goes, maybe this will already suffice.

As to the naming, if ever, I thought `excludeChunkFromExtractor`, although specific, still did all that its name suggests to do. `client:false` wouldn't be true as this is tied to our `LazyRender` component because all we want to do is not append the `<script>` tag related to the chunk on SSR. For `track:false`, not sure if I'd understand it.

I'll check out your suggestion, perhaps we won't need to add this option after all.

pajaron-mms:
@theKashey So I've done:

```
const extractor = new ChunkExtractor({
  stats: {},
})
```

Not sure if I need to add more dummy data there. The Chunk is successfully not appended with SSR, however when the component does come into view I get the error:

```
loadable-components: failed to synchronously load component, which expected to be available
```

theKashey:
Not sure what you did, the idea was to:
- keep top level extractor + manager
- create one more extractor + manager
- wrap "non-trackable" locations with the second one, sending results only from the first one to the client.

pajaron-mms:
Yes, this is what I did.
- I didn't touch the top level extractor + manager
- created one more extractor with `stats: {}` on the component where I'm using the loadable component
- wrapped the target loadable component with the new manager

I've excluded using our `LazyRender` component to rule it out. And so now we get this error on page-load instead of when scrolling the component into view.

If you say this should work, perhaps it's in our setup that I need to debug the cause of the error.

pajaron-mms:
when the error happens:
```
// createLoadable.js
try {
          var loadedModule = ctor.requireSync(this.props); // loadedModule = undefined

pajaron-mms:
Actually It seems it was a problem on our side! It seems to be working now! Thanks for the help.

theKashey:
Actually adding per component "track:false" option still seems to be a good idea, as well as documenting the way it might work.
Feel free to update and reopen this PR (but I will ask for documentation for the new field as well)

pajaron-mms:
Cool, will do that then!

pajaron-mms:
what about `dontTrack: true`? doing `track: false` would mean I'd need to change more and default it to `false` when the current default is `options = {}`

theKashey:
Oh, that's a complex answer. While using "false" options is usually an antipattern, I still like how declarative they are.
We already have `ssr:false` -> and it "sounds" right, "dontSSR:true" is completely meaningless.

`dontTrack` is better than `dontSSR`, but already has other meaning (do-not-track header), and not using other well known _symbol_ is quite important for me.

fresk-nc:
What about `lazy: true`?

pajaron-mms:
Might need to revive to have a cleaner code from our side.

PR #768, opened 2021-05-12
Make example OS agnostic

GeorgeTaveras1231:
## Summary

This changes the example in server docs to be compatible with Unix and Windows.

## Test plan
N/A


theKashey:
👍

PR #778, opened 2021-05-26
feat: add stats option to webpack plugin

### LINKED ISSUE #770: Missing assets from loadable-stats.json
## 🐛 Bug Report
We are receiving the following error when generating HTML via our server-side-rendering process. 

```
  const nodeStats = path.join(process.cwd(), 'build/node/loadable-stats.json')
  const webStats = path.join(process.cwd(), 'build/loadable-stats.json')

  const nodeExtractor = new ChunkExtractor({
    statsFile: nodeStats, outputPath: path.resolve(process.cwd(), 'build/node')
  })
  const { default: App, React, renderToString } = nodeExtractor.requireEntrypoint()
  const webExtractor = new ChunkExtractor({
    statsFile: webStats, outputPath: path.resolve(process.cwd(), 'build/web')
  })

  ```

Here's the error:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
    at validateString (internal/validators.js:112:11)
    at Object.extname (path.js:1231:5)
    at getFileScriptType (/Users/noah/work/lib/node_modules/@loadable/server/lib/ChunkExtractor.js:49:60)
    at isScriptFile (/Users/noah/work/lib/node_modules/@loadable/server/lib/ChunkExtractor.js:53:10)
    at /Users/noah/work/lib/node_modules/@loadable/server/lib/ChunkExtractor.js:366:11
```

It appears that the cause is related to the way the assets within the `loadable-stats.json` block is generated. 

Here's the loadable-stats.json using 5.15

```
{
  "hash": "aeebda487075c80e6709",
  "publicPath": "/",
  "outputPath": "/Users/nmarkon/work/content-site-connect2local/build/node",
  "assetsByChunkName": {
   ...snip..
  },
  "assets": [
    {
      "type": "assets by status",
      "cached": true,
      "filteredChildren": 70,
      "size": 281927
    }
  ],
```

In `5.14.2` the assets block in contained information regarding each chunk such as:

```
{
  "hash": "4ee6a975c8dd4cdbc020",
  "version": "5.37.0",
  "time": null,
  "publicPath": "/",
  "outputPath": "/Users/noah/work/lib/build/node",
  "assetsByChunkName": {
   .... snip....
  },
  "assets": [
    {
      "type": "asset",
      "name": "static/js/name-goes-here.d2b992cb.chunk.js",
      "size": 3518,
      "emitted": false,
      "comparedForEmit": false,
      "cached": true,
      "info": {
        "immutable": true,
        "contenthash": "d2b992cb",
        "javascriptModule": false,
        "minimized": true
      },
      "chunkNames": [
        "name-goes-here"
      ],
      "chunkIdHints": [],
      "auxiliaryChunkNames": [],
      "auxiliaryChunkIdHints": [],
      "related": {},
      "chunks": [
        4677
      ],
      "auxiliaryChunks": [],
      "isOverSizeLimit": false
    },
...snip... 
```
It looks like this change is the root cause:
https://github.com/gregberge/loadable-components/commit/9ad6bf569a28ff553a5adf5104b55c6dcba47016

It appears that the `all: false` is preventing the cached assets from being added. If `cachedAssets: true` is added to the JSON then the cached assets are added as expected.

depoulo:
## Summary

Fix #770 as per [my suggestion](https://github.com/gregberge/loadable-components/issues/770#issuecomment-847797995).

## Test plan

1. Use plugin as before. Output should be identical.
2. Use plugin while passing custom `stats` options. The generated stats JSON file matches those options.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -11,35 +11,45 @@ class LoadablePlugin {
     writeToDisk,
     outputAsset = true,
     chunkLoadingGlobal = '__LOADABLE_LOADED_CHUNKS__',
+    stats,
   } = {}) {
-    this.opts = { filename, writeToDisk, outputAsset, path, chunkLoadingGlobal }
+    this.opts = {
+      filename,
+      writeToDisk,
+      outputAsset,
+      path,
+      chunkLoadingGlobal,
+      stats,
+    }
 
     // The Webpack compiler instance
     this.compiler = null
   }
 
   handleEmit = compilation => {
-    const stats = compilation.getStats().toJson({
-      all: false,
-      assets: true,
-      chunks: false,
-      chunkGroups: true,
-      chunkGroupChildren: true,
-      hash: true,
-      ids: true,
-      outputPath: true,
-      publicPath: true,
-    })
+    const stats = compilation.getStats().toJson(
+      this.opts.stats || {
+        all: false,
+        assets: true,
+        chunks: false,
+        chunkGroups: true,
+        chunkGroupChildren: true,
+        hash: true,
+        ids: true,
+        outputPath: true,
+        publicPath: true,
+      },
+    )
 
     stats.generator = 'loadable-components'
 
     // we don't need all chunk information, only a type
-    stats.chunks = [...compilation.chunks].map((chunk) => {
+    stats.chunks = [...compilation.chunks].map(chunk => {
       return {
         id: chunk.id,
         files: [...chunk.files],
-      };
-    });
+      }
+    })
 
     const result = JSON.stringify(stats, null, 2)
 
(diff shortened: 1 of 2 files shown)


tomkelsey:
Hey,

I've read the related conversation to this PR and saw (https://github.com/gregberge/loadable-components/issues/770#issuecomment-848551220) you say perhaps this was overkill and that #779 would suffice.

Personally, I re-use the assets json file for a few other bits and bobs and so the most recent update to loadable-components broke my setup. I've had to patch the code to add in an extra stats option (namely `entrypoints`) so this PR - allowing people to customise the stats options - would still be valuable for my use case 👍

theKashey:
Actually I always thought about splitting plugin into two parts - Webpack and stats.

you should not use plugin output for anything except loadable itself, please try to learn from this incident...
But nothing is stopping loadable to derive stats it needs from “your stats”

that automatically “grants” you ability to use any plugin you want and decouples loadable server from Webpack integration.

Issue #769, opened 2021-05-13
Expose just scripts for chunks


LorenDorez:
## 🚀 Feature Proposal

I would like a method on the extract to get the scripts minus the entry points.

## Motivation

Since im doing SSR through ReactJS.net i already have the entry scripts load on the page. I just need the chunks.

## Example

Current
```
 getMainAssets(scriptType) {
    const chunks = [...this.entrypoints, ...this.chunks]
    const assets = this.getChunkAssets(chunks)
    if (scriptType) {
      return assets.filter(asset => asset.scriptType === scriptType)
    }
    return assets
  }

  getScriptTags(extraProps = {}) {
    const requiredScriptTag = this.getRequiredChunksScriptTag(extraProps)
    const mainAssets = this.getMainAssets('script')
    const assetsScriptTags = mainAssets.map(asset =>
      assetToScriptTag(asset, extraProps),
    )
    return joinTags([requiredScriptTag, ...assetsScriptTags])
  }
```

Suggested quick mockup
```
 getChunkAssets(scriptType) {
    const chunks = [...this.chunks]
    const assets = this.getChunkAssets(chunks)
    if (scriptType) {
      return assets.filter(asset => asset.scriptType === scriptType)
    }
    return assets
  }

  getChunkScriptTags(extraProps = {}) {
    const requiredScriptTag = this.getRequiredChunksScriptTag(extraProps)
    const mainAssets = this.getMainAssets('script')
    const assetsScriptTags = mainAssets.map(asset =>
      assetToScriptTag(asset, extraProps),
    )
    return joinTags([requiredScriptTag, ...assetsScriptTags])
  }
```

## Pitch

This will allow for better integration with a Popular SSR library for react called ReactJS.net

theKashey:
Sound right. It's time to "DRY" this package a little and provide more fine-grained functionality.

charle692:
So I have a similar need but instead of it being to exclude **entrypoints** I want to be able to specify which chunk I want to generate script tags/elements or assets from. I feel like we could solve both of these issues by adding the ability to specify which chunks you want to generate elements/tags for.

So for example, within the **ChunkExtractor** I want to be able to specify multiple entrypoints as below:
```ts
const extractor = new ChunkExtractor({
  statsFile: statsFilePath,
  entrypoints: ["polyfill", "foobar"],
});
```

[...]

PR #784, opened 2021-05-31
Added getScriptTagsByChunks and getScriptElementsByChunks

charle692:
## Summary

This PR aims to solve #769 by providing an API where you can generate script tags/elements for a specific chunk excluding other non-specified chunks. 

## Test plan

1. All other APIs will work as before, they remain unchanged.
2. For the new APIs, the arguments is now a single object with optional properties. 
  - `extraProps` works exactly as before either an object or a function
  - `includedChunks` allows you to specify which chunks you want to generate tags/elements for
3. You can see an example usage in the specs that I've added, hopefully those are enough to demonstrate how this works

## Notes

I'm open to discussion with the API design here, let me know if there's anything that you would like changed or designer another way.

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -267,6 +267,14 @@ class ChunkExtractor {
     return one(chunks)
   }
 
+  getChunkAssetsByType(scriptType, chunks) {
+    const assets = this.getChunkAssets(chunks)
+    if (scriptType) {
+      return assets.filter(asset => asset.scriptType === scriptType)
+    }
+    return assets
+  }
+
   getChunkChildAssets(chunks, type) {
     const one = chunk => {
       const chunkGroup = this.getChunkGroup(chunk)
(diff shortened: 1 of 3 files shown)


fresk-nc:
Good job, but what about `getStyleElements` and `getLinkElements`?

fresk-nc:
https://github.com/gregberge/loadable-components/pull/749

charle692:
> Good job, but what about `getStyleElementsByChunks`, `getStyleTagsByChunks`, `getLinkElementsByChunks` and `getLinkTagsByChunks`? Maybe we should add `includedChunks` to the `collectChunks`?

Sure, I mean we can consider adding those the style and link related functions. In my particular use case I don't have a need and I also tried to think of any cases where the others might be required but couldn't come up with something. So I didn't include them. 

Obviously that doesn't mean there's no one else that might not benefit from them. I'm open to adding them to this if we want, or I can also create a subsequent PR if we want to keep this one scoped to just scripts.

charle692:
Bump

theKashey:
Hey. Sorry for so late reply, got a little overwhelmed with notifications and found this one only now.

There are a few things:
- I generally agree that such functionality can be very useful.
- However such functionality should be built aside, not a part of the server package
- There is ongoing work to extract all `webpack-plugin` related logic to another package sitting in-between `server` and `webpack-plugin`, and decoupling these two from each other

Once the initial scaffolding will be complete I would probably retrofit this PR, especially tests, to that new entity.

charle692:
Ok yeah that works for me, once the new package is ready to go, I will add this functionality there. Right now we can't upgrade until this is merged, we are experiencing slow downs with webpack v5 that were fixed in v5.15.0 but require functionality that was also removed by that release.

Is there somewhere I should keep looking to see once the new package is ready?

Issue #563, opened 2020-04-27
integrity isn't added to assets in loadable-stats.json


abnud1:
## 🐛 Bug Report
I'm having the same problem as this issue: https://github.com/gregberge/loadable-components/issues/344

However a PR Was merged which solved this issue and in the past it worked, for some reason it now
doesn't work at all.
## To Reproduce

1- install webpack-subresource-plugin and setup loadable components for SSR(webpack-plugin et all)
2- add the plugins in webpack config like this:
```
            new SriPlugin({
                hashFuncNames: ["sha256"],
                enabled: !isDevelopment,
            }),
            new LoadablePlugin(),
```
also add `crossOriginLoading: "anonymous"` to `webpack.output`.
3- build the bundle, in the generted loadable-stats.json you'll see no integrity attribute.
4- in the server bundle call `chunkExtractor.getScriptTags({ crossorigin: "anonymous" })`, you'll notice no integrity attribute
## Expected behavior
`chunkExtractor.getScriptTags()`, `chunkExtractor.getStyleTags()`,
`chunkExtractor.getLinkTags()` would set integrity attribute.

`npx envinfo --system --binaries --npmPackages @loadable/component,@loadable/server,@loadable/webpack-plugin,@loadable/babel-plugin --markdown --clipboard` output:
```
## System:
 - OS: Windows 10 10.0.18363
 - CPU: (8) x64 Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
 - Memory: 1.27 GB / 7.89 GB
## Binaries:
 - Node: 14.0.0 - C:\Program Files\nodejs\node.EXE
 - npm: 6.14.4 - ~\AppData\Roaming\npm\npm.CMD
```
Other Information:
webpack version: 4.43
loadable-components version: 5.12
webpack-subresource-plugin version: 1.4.0

jarrett-confrey:
@abnud1 i am seeing this issue as well. were you able to figure out why this happens?
```
## System:
 - OS: macOS 11.4
 - CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 - Memory: 61.25 MB / 16.00 GB
 - Shell: 5.8 - /bin/zsh
## Binaries:
 - Node: 14.16.0 - /usr/local/bin/node
 - Yarn: 1.22.10 - ~/my-app/node_modules/.bin/yarn
 - npm: 6.14.11 - /usr/local/bin/npm
[...]

PR #803, opened 2021-07-22
fix: subresource integrity

jarrett-confrey:
## Summary
this pull request solves the problem of integrity not being added to script tags when using [`webpack-subresource-integrity`](https://github.com/waysact/webpack-subresource-integrity). it seems that either https://github.com/gregberge/loadable-components/pull/436 did not include everything necessary or possibly my configuration is different but i was unable to get things working properly without these changes. other users are seeing this issue as well as seen in https://github.com/gregberge/loadable-components/issues/563

## Test plan
- add [`webpack-subresource-integrity`](https://github.com/waysact/webpack-subresource-integrity) to your webpack plugins
- you should now see the `integrity` attribute in `loadable-stats.json` and on your `script` tags when using `ChunkExtractor.getScriptTags`

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -39,8 +39,20 @@ class LoadablePlugin {
       return {
         id: chunk.id,
         files: [...chunk.files],
-      };
-    });
+      }
+    })
+
+    // update namedChunkGroups with integrity from webpack-subresource-integrity if available
+    Object.values(stats.namedChunkGroups).forEach(namedChunkGroup => {
+      namedChunkGroup.assets.forEach(namedChunkGroupAsset => {
+        if (!namedChunkGroupAsset.integrity) {
+          const asset = stats.assets.find(a => a.name === namedChunkGroupAsset.name) || {}
+          if (asset.integrity) {
+            namedChunkGroupAsset.integrity = asset.integrity
+          }
+        }
+      })
+    })
 
     const result = JSON.stringify(stats, null, 2)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,6 +64,7 @@ _Before_ submitting a pull request, please make sure the following is done…
 8.  Ensure the test suite passes via `yarn test`.
 
     ```sh-session
+    $ yarn test:prepare # build example and generate fixtures before running tests
     $ yarn test
     ```
 
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -233,9 +233,12 @@ class ChunkExtractor {
   createChunkAsset({ filename, chunk, type, linkType }) {
     const resolvedFilename =
       typeof filename === 'object' && filename.name ? filename.name : filename
+    const resolvedIntegrity =
+      typeof filename === 'object' && filename.integrity ? filename.integrity : null
 
     return {
       filename: resolvedFilename,
+      integrity: resolvedIntegrity,
       scriptType: getFileScriptType(resolvedFilename),
       chunk,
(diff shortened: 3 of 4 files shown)


packages/webpack-plugin/src/index.js
```diff
@@ -123,6 +123,25 @@ class LoadablePlugin {
             },
           )
         })
+
+        // update loadable-stats.json with integrity from webpack-subresource-integrity if available
+        compiler.hooks.done.tap(name, stats => {
+          const statsAssets = stats.toJson({ all: false, assets: true }).assets
+          const outputFolder = this.compiler.options.output.path
+          const outputFile = nodePath.resolve(outputFolder, this.opts.filename)
+          const fileContents = fs.readFileSync(outputFile)
```
theKashey, line 132: this fill will exists only if `writeToDisk` is enabled

packages/webpack-plugin/src/index.js
```diff
@@ -123,6 +123,25 @@ class LoadablePlugin {
...
+          Object.values(loadableStats.namedChunkGroups).forEach(namedChunkGroup => {
+            namedChunkGroup.assets.forEach(namedChunkGroupAsset => {
+              const statsAsset = statsAssets.find(asset => asset.name === namedChunkGroupAsset.name)
+              if (typeof statsAsset === 'object' && statsAsset.integrity) {
+                namedChunkGroupAsset.integrity = statsAsset.integrity
+              }
+            })
+          })
+          const loadableStatsJson = JSON.stringify(loadableStats, null, 2)
+          fs.writeFileSync(outputFile, loadableStatsJson)
```
theKashey, line 143: There is `writeAssetsFile` which does the same

packages/webpack-plugin/src/index.js
```diff
@@ -123,6 +123,25 @@ class LoadablePlugin {
             },
           )
         })
+
+        // update loadable-stats.json with integrity from webpack-subresource-integrity if available
+        compiler.hooks.done.tap(name, stats => {
+          const statsAssets = stats.toJson({ all: false, assets: true }).assets
```
theKashey, line 129: this line does the same as line 22 of the current file. 
There should be a way to run operation only once.

packages/webpack-plugin/src/index.js
```diff
@@ -123,6 +123,25 @@ class LoadablePlugin {
             },
           )
         })
+
```
theKashey, line 126: checked - SriPlugin extends assets information at PROCESS_ASSETS_STAGE_OPTIMIZE_HASH stage, so at PROCESS_ASSETS_STAGE_REPORT all extra fields should be already a part of stats.
- https://webpack.js.org/api/compilation-hooks/
- https://github.com/waysact/webpack-subresource-integrity/blob/master/index.js#L396

theKashey (changes requested):
There should be close to zero changes around `LoadablePlugin.apply`

packages/webpack-plugin/src/index.js
jarrett-confrey: youre right, thanks! i wasnt seeing the integrity available at this stage before but i must have been looking in the wrong place at the time

jarrett-confrey:
> There should be close to zero changes around `LoadablePlugin.apply`

done, thanks for the feedback!

theKashey (commented):
Wow it's really simple now! I'll review it shortly

jarrett-confrey:
hey @theKashey, just checking if youve had a chance to take another look? currently i wrote a new plugin to do this work for our project but i would prefer not to keep it around. thanks!

theKashey:
Verified and merged. 👍

jarrett-confrey:
thank you @theKashey! any chance for a new release with this in it?

PR #828, opened 2021-09-03
Fix getInlineStyleTags/Elements docs.

PatrickNausha:
## Summary

These docs disagree with their implementation. The docs mention `link` tags but the implementation uses `style` tags. The implementation seems obviously correct so I assume it's the docs that need fixing.

Here is `getInlineStyleTags` using `style` tags as expected:
https://github.com/gregberge/loadable-components/blob/629c80e42eea605a32dafd7b52faf3140e1aab28/packages/server/src/ChunkExtractor.test.js#L463-L479

Here is `getInlineStyleElements` using `style` tags as expected:
https://github.com/gregberge/loadable-components/blob/629c80e42eea605a32dafd7b52faf3140e1aab28/packages/server/src/ChunkExtractor.test.js#L622-L650

## Test plan

N/A

### DIFF
--- a/website/src/pages/docs/api-loadable-server.mdx
+++ b/website/src/pages/docs/api-loadable-server.mdx
@@ -132,7 +132,7 @@ const head = renderToString(<head>{chunkExtractor.getStyleElements()}</head>)
 
 ## chunkExtractor.getInlineStyleTags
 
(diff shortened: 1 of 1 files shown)


theKashey:
👍 yep, there is no such thing as `inline link tag`

PR #830, opened 2021-09-15
Cache promise on load/preload

macko911:
## Summary

I guess the cache in `preload` and `load` methods is not used on purpose but I was curious whether you'd be interested in this change, maybe under an option or a different static method `cache`.

The use case for me is that I did try to prefetch pages on link hover, but after transiting to the prefetched page I saw a blink of white page in between the route transitions, even though I was blocking the transition until the module was fetched.

This only happened on the first transition, subsequent transitions to the cached pages didn't have the blink anymore, which lead me to this change.

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -343,10 +343,13 @@ function createLoadable({
 
     // In future, preload could use `<link rel="preload">`
     Loadable.preload = props => {
-      ctor.requireAsync(props)
+      cache[getCacheKey()] = ctor.requireAsync(props);
     }
 
-    Loadable.load = props => ctor.requireAsync(props)
+    Loadable.load = props => {
+      cache[getCacheKey()] = ctor.requireAsync(props);
+      return cache[getCacheKey()];
+    }
 
     return Loadable
   }


packages/component/src/createLoadable.js
```diff
@@ -343,10 +343,13 @@ function createLoadable({
 
     // In future, preload could use `<link rel="preload">`
     Loadable.preload = props => {
-      ctor.requireAsync(props)
+      cache[getCacheKey()] = ctor.requireAsync(props);
     }
 
-    Loadable.load = props => ctor.requireAsync(props)
+    Loadable.load = props => {
+      cache[getCacheKey()] = ctor.requireAsync(props);
```
theKashey, line 350: I think the right way forward is to (re)use ‘ resolveAsync’ here, but it does mean that you are right 👍 , it’s all about caches.

cecchi:
Hey @theKashey -- is this ready to go? Not sure I understood the review comment, but happy to help with implementation if I can. We're running into this issue as well.

theKashey:
Everything is fine, I am just have developed a habit to always seek for perfection as mentoring our devs is a big part of my job. 
And that is not the best habit.

gonna merge this PR and apply changes I would like to see by myself, that’s just a few lines change.

theKashey:
Everything is fine, I am just have developed a habit to always seek for perfection as mentoring our devs is a big part of my job. 
And that is not the best habit.

gonna merge this PR and apply changes I would like to see by myself, that’s just a few lines change.

Issue #821, opened 2021-08-29
Stale `statsFile`


Dattaya:
## 🐛 Bug Report

I hot reload the server in my boilerplate and for some reason `new ChunkExtractor({ statsFile })` loads old `statsFile` and it's not present in `require.cache` so `clearModuleCache` can't do anything much. Changing `smartRequire` in utils to `JSON.parse(fs.readFileSync(modulePath));` fixes the problem. I can send a PR if you're interested.

## To Reproduce

Tested on Linux, might not work on other OSs:
- `git clone https://github.com/Dattaya/ts-react-ssr-basis && cd ts-react-ssr-basis && npm i && npm start`
- Open `http://localhost:3000/` and view page source (Ctrl+U in Chrome)
- Modify `common/components/App/index.tsx`, save it, reload the page source and observe that `script src=` points to the old js file.

## Expected behavior

`ChunkExtractor.getScriptTags` returns updated assets.  

## `npx envinfo`

```bash
## System:
 - OS: Linux 5.11 Ubuntu 20.04.3 LTS (Focal Fossa)
 - Shell: 5.0.17 - /bin/bash
## Binaries:
 - Node: 14.15.3 - ~/.nvm/versions/node/v14.15.3/bin/node
 - Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.3/bin/yarn
 - npm: 6.14.11 - ~/.nvm/versions/node/v14.15.3/bin/npm
## npmPackages:
 - @loadable/babel-plugin: ^5.13.2 => 5.13.2 
 - @loadable/component: ^5.15.0 => 5.15.0 
 - @loadable/server: ^5.15.1 => 5.15.1 
 - @loadable/webpack-plugin: ^5.15.1 => 5.15.1
```

rtymchyk:
@Dattaya What was the resolution here?

Dattaya:
@rtymchyk, sorry for the delayed response. I changed my mind, because usually issues are closed when they're fixed, not when included in a new release. In the meantime you can modify `util.js` in `node_modules` manually.

PR #838, opened 2021-09-23
fix: Clear correct cache when compiling with Webpack

Mordred:
## Summary

Using `module.hot` is forcing me to use Webpack `HotModuleReplacementPlugin` and excluding `@loadable/server` from `nodeExternal()` so that `module.hot` is present. (8f10896058cdfd542c2aca1a982c2e14b1d9707a)

Then `__non_webpack_require__` is used in `smartRequire` (https://github.com/gregberge/loadable-components/blob/main/packages/server/src/util.js#L30) but `clearModuleCache` is clearing cache from `require` not from `__non_webpack_require__`.

This PR should fix this problem.

## Test plan

See #821

### DIFF
--- a/packages/server/src/util.js
+++ b/packages/server/src/util.js
@@ -1,5 +1,11 @@
+// Use __non_webpack_require__ to prevent Webpack from compiling it
+// when the server-side code is compiled with Webpack
+// eslint-disable-next-line camelcase, no-undef, global-require, import/no-dynamic-require, no-eval
+const getRequire = () => typeof __non_webpack_require__ !== 'undefined' ? __non_webpack_require__ : eval('require');
+
 export const clearModuleCache = moduleName => {
-  const m = require.cache[moduleName]
+  const { cache } = getRequire();
+  const m = cache[moduleName]
   if (m) {
     // remove self from own parents
     if (m.parent && m.parent.children) {
@@ -13,7 +19,7 @@ export const clearModuleCache = moduleName => {
         }
       })
     }
-    delete require.cache[moduleName]
+    delete cache[moduleName]
   }
 }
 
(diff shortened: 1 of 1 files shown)


theKashey:
Sorry, took some time to connect all the dots and pieces and understand how it's working, or how it's not working :)

You are absolutely correct, `smartRequire` and `clearCache` should use the same "source" for module information.

packages/server/src/util.js
```diff
@@ -1,5 +1,6 @@
 export const clearModuleCache = moduleName => {
-  const m = require.cache[moduleName]
+  const cache = (typeof __non_webpack_require__ !== 'undefined' ? __non_webpack_require__ : require).cache;
```
theKashey, line 2: Sorry for being picky, but let's safe future developers time and make it right this time

Let's use "single source" for both operations if the HAVE to be synchronized.
```js
const getRequire = () => typeof __non_webpack_require__ !== 'undefined' ? __non_webpack_require__ : require;

export const clearModuleCache = moduleName => {
  const cache = getRequire().cache; // 👈
...
}

export const smartRequire = modulePath => {
  if (process.env.NODE_ENV !== 'production' && module.hot) {
    clearModuleCache(modulePath)
  }
 
  return getRequire(moduleName);  // 👈
}
```

theKashey (approved):
The approach is correct. Let's make it ideal

Mordred:
@theKashey OK, no problem. I've refactored it based on your suggestions and kept `eval` - not sure if it is necessary, but it was there in original code.

packages/server/src/util.js
```diff
@@ -22,16 +28,7 @@ export const smartRequire = modulePath => {
...
-  // when the server-side code is compiled with Webpack
-  // eslint-disable-next-line camelcase
-  if (typeof __non_webpack_require__ !== 'undefined') {
-    // eslint-disable-next-line no-undef
-    return __non_webpack_require__(modulePath)
-  }
-
-  // eslint-disable-next-line global-require, import/no-dynamic-require, no-eval
-  return eval('require')(modulePath)
+  return getRequire(modulePath)
```
theKashey, line 31:
```suggestion
  return getRequire()(modulePath)
```

theKashey:
thank you for your contribution, I will release the update shortly.

kiyasov:
> thank you for your contribution, I will release the update shortly.

Hi! When will it be released?

rtymchyk:
Would appreciate a release with this commit when possible 🙏

rtymchyk:
🦗

theKashey:
🙄😅

theKashey:
https://github.com/gregberge/loadable-components/releases/tag/v5.15.2

PR #839, opened 2021-09-24
fix: loadAsync resolve

jinasonlin:
## Summary

I think loadAsync should like loadSync resolve module and copy Component statics (hoistNonReactStatics).

In some cases, I hope loadable can trigger load end, so add onLoad in options.

PR #847, opened 2021-10-18
feat: add onLoad in options

jinasonlin:
## Summary

Add `onLoad` in options. When async component load finish, `onLoad` will be trigger.

## Test plan

Previous comments #839


theKashey:
There is a little question to answer - when to call `onLoad` - on the first load, or on the every "load"(even if it did not happen)

jinasonlin:
I expect that is on the first load, like the library's `onload`.

Issue #853, opened 2021-11-17
__LOADABLE_REQUIRED_CHUNKS__ empty 


KERIST:
## 💬 Questions and Help

Loadable Components project is young, but please before asking your question:

- Read carefully the README of the project
- Search if your answer has already been answered in old issues

After you can submit your question and we will be happy to help you!

Hello. The problem is __LOADABLE_REQUIRED_CHUNKS__ is empty and because of it I suppose react don't hydrate and app is running with no event listeners.  I've been trying to solve it for many days. There is a [similiar](https://github.com/gregberge/loadable-components/issues/312) issue, but it closed unresolved. Please help me.

Here is a repository with code: https://github.com/KERIST/loadable-help.

LudvigHz:
Hey @KERIST. I had the exact same issue. It took me some time to realize as well, but you have to render the app _before_ you collect the script tags. This is how loadable knows what chunks are required for the current render.

I.e. here:
https://github.com/KERIST/loadable-help/blob/25f4957651447204c94e1ad0c2e64966b63d4f5a/server/index.js#L42-L59

Move your `renderToString` to above the `getLinkTags` etc.
```diff
+ const appHtml = ReactDOMServer.renderToString(sheet.collectStyles(jsx));

  const scriptTags = webExtractor.getScriptTags();
  const linkTags = webExtractor.getLinkTags();
  const styleTags = webExtractor.getStyleTags();

- const appHtml = ReactDOMServer.renderToString(sheet.collectStyles(jsx));
```

It's documented [here](https://loadable-components.com/docs/server-side-rendering/#3-setup-chunkextractor-server-side) but it **really** should explicitly say that the order is very important.

KERIST:
@LudvigHz Thank you so much for your help!

PR #855, opened 2021-12-01
 Add note on order of calls when server-side rendering

LudvigHz:
## Summary

The docs on server-side rendering could be interpreted incorrectly and one can miss the important fact that the order of calls to `ChunkExtractor` matters alot. Specifically that you have to render the app _before_ collecting script tags.

This adds a note that explicitly states the importance of the call order. I personally spent way too many hours debugging why my app was not hydrating properly, and without any errors or docs on this matter, it's quite easy to miss this important fact.

There are also a number of issues where the cause is this exact configuration issue. Hopefully this will help alleviate any similar mistakes in the future.

### DIFF
--- a/website/src/pages/docs/server-side-rendering.mdx
+++ b/website/src/pages/docs/server-side-rendering.mdx
@@ -56,6 +56,10 @@ const jsx = extractor.collectChunks(<YourApp />)
 // Render your application
 const html = renderToString(jsx)
 
+/* `renderToString` must be called before `getScriptTags`, `getLinkTags`
+ * and `getStyleTags`.
+ */
+
 // You can now collect your script tags
 const scriptTags = extractor.getScriptTags() // or extractor.getScriptElements();
 


website/src/pages/docs/server-side-rendering.mdx
```diff
@@ -66,6 +66,10 @@ const linkTags = extractor.getLinkTags() // or extractor.getLinkElements();
 const styleTags = extractor.getStyleTags() // or extractor.getStyleElements();
 ```
 
+> Note that you must call `renderToString` before collecting the script tags. Otherwise loadable
```
theKashey, line 69: in the given example `collectChunks` is used before `renderToString`. 
Sorry, but I am not sure that this note would be helpful and not cause more confusion.

LudvigHz: Of course, I'm referring to `getScriptTags`, `getLinkTags` and `getStyleTags`. I could maybe be even more explicit so there is no room for confusipn?

I personally spent over a day hitting my head with this issue, and there are issues that report the same confusion (#853 is one). I think a note like this will help a lot of people avoid this issue and better understand how to use loadable.

theKashey: Can a note right after `renderToString`, ie the one "in the example" do a better job?

LudvigHz: Sure. I added a note in the code listing now, that hopefully is clearer and less obtrusive.

theKashey (approved):

theKashey:
👍

PR #872, opened 2022-02-07
feat: add transformStats option

shredor:
## Summary
Add `transformStats` option to webpack plugin. 
Motivation: sometimes you need to add additional info to loadable stats json or remove some fields. For example if you don't use `extractor.requireEntrypoint` method you can remove `stats` field and reduce stats file size several times.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -11,15 +11,23 @@ class LoadablePlugin {
     writeToDisk,
     outputAsset = true,
     chunkLoadingGlobal = '__LOADABLE_LOADED_CHUNKS__',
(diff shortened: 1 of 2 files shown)


theKashey:
In general, it is a bad idea to let others mutate some source they should not be aware of. 
Currently, loadable stat file is a direct translation of webpack stats, but it can change in the future, and having imperative transformations at the clientside will make such migration complicated.

Sorry, I cannot accept this change to a public API.

If you are looking to make the stat file more compact - let's discuss how it can be made, when the trimmed result is consumable and understand which other journey will be broken.
The outcome can be documented and implemented as different "modes" or different versions of webpack-plugin.

Issue #640, opened 2020-10-14
loadable not working in SSR with webpack module federation - 


ajayjaggi:
## 🐛 Bug Report

loadable-components: failed to synchronously load component, which expected to be available { fileName: './src/shared/dedicated/index.js',
  chunkName: 'Dedicated',
  error: 'Cannot read property \'call\' of undefined' }
(node:7562) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'call' of undefined
    at __webpack_require__ (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/main.js:394:42)
    at Module../src/shared/components/Footer/index.js (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/Dedicated.server.js:21:71)
    at __webpack_require__ (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/main.js:394:42)
    at Module../src/shared/dedicated/index.js (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/Dedicated.server.js:60:76)
    at __webpack_require__ (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/main.js:394:42)
    at Object.requireSync (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/src_server_render_js.server.js:225:14)
    at InnerLoadable.loadSync (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/vendors-node_modules_loadable_server_lib_index_js-node_modules_express_index_js-node_modules_-ee7ccd.js:420:35)
    at new InnerLoadable (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/vendors-node_modules_loadable_server_lib_index_js-node_modules_express_index_js-node_modules_-ee7ccd.js:315:17)
    at processChild (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/vendors-node_modules_loadable_server_lib_index_js-node_modules_express_index_js-node_modules_-ee7ccd.js:56603:14)
    at resolve (/Users/ajay/Desktop/WebStrom/MicroFrontend/Edge/dist/server/vendors-node_modules_loadable_server_lib_index_js-node_modules_express_index_js-node_modules_-ee7ccd.js:56568:5)
[...]

PR #883, opened 2022-03-17
WIP: Fix SSR by adding unused synchronous import for every loadable call

IMalyugin:
Simply add that unused synchronous import for every loadable call to fix SSR

## Summary

Dirty way of fixing the issue with SSR only working with WMF if every loadable call is backed by a sync call;
For details see: https://github.com/gregberge/loadable-components/issues/640#issuecomment-1068800491

### DIFF
--- a/packages/babel-plugin/src/index.js
+++ b/packages/babel-plugin/src/index.js
@@ -104,6 +104,15 @@ const loadablePlugin = api => {
     } else {
       funcPath.replaceWith(object)
     }
+
+    /** Append unused synchronous import for WMF  */
+    const importPath = callPath.node.arguments[0].value;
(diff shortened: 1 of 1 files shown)


IMalyugin:
The code works, but:
1. it's only tested for normal imports with no params;
2. It is probably a bad idea to have babelPlugin with such a hack for client build
3. For simplicity the whole solution is inlined in transformImport

theKashey:
Adding a `static import` will completely disable code-splitting. You cannot do this.

IMalyugin:
Updated SSR to exclude static import from client-side.

IMalyugin:
Closed in favour of https://github.com/gregberge/loadable-components/pull/885

PR #885, opened 2022-03-21
WIP: Fix SSR and CSR for federated modules

IMalyugin:
## Changes

- add loadableModulesMap to register and preload modules for client- and server-side rendering;
- loadableReady now calls importAsync for all the federated modules (listed in named section) and only resolves once that import finishes;
- loadableReady now runs on the server, but instead of preloading required modules, it preloads all the federated modules. During that preload, new federated modules may be discovered, so the process continues till the entire chain is required;

## Summary

Fixes SSR and CSR for ModuleFederation.
For details see: https://github.com/gregberge/loadable-components/issues/640

There are currently issues both server- and client-side:
- loadable currently fails server-side unless static import for each of the federated modules exists. That hack allows wmf to preload these modules during async bootstrap.
- loadableReady call on the client currently breaks on the client, because requireSync expects loadable chunks to be available on the page. Example currently hydrate the page without loadableReady, which results in loadable running in async mode, and can cause side effects, if modules aren't available on the page;

This PR is meant to fix both these problems,

## Test plan

Unsolved major issues:
- [ ] Eager static imports behind async boundary are not added to loadable manifest
- [ ] No way to mark async boundary import as required loadable dependency

Minor issues that need to be addressed:
- [x] Need to rebuild loadable and see if it works, might be some side effects;
- [ ] { ssr: false } flag might not be supported;
- [ ] There is no warning for not providing federated chunks synchronously;
- [ ] Only tested for webpack target: 'web' | 'node';
- [ ] Not tested with fully dynamic components (they are probably not going to be supported);
- [ ] Not tested with loadable.lib;
- [ ] LoadableReady used to support synchronous call, now its promise-based. We might want to make it work synchronously again for backwards compatibility;
- [ ] Check if this is working with production builds;
- [ ] Check if client-side only required components are preloaded;

### DIFF
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -5,7 +5,7 @@ import hoistNonReactStatics from 'hoist-non-react-statics'
 import { invariant } from './util'
 import Context from './Context'
 import { LOADABLE_SHARED } from './shared'
-import { register } from './loadableModulesMap';
+import { register } from './federatedResolver';
 
 const STATUS_PENDING = 'PENDING'
 const STATUS_RESOLVED = 'RESOLVED'
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -5,6 +5,7 @@ import hoistNonReactStatics from 'hoist-non-react-statics'
 import { invariant } from './util'
 import Context from './Context'
 import { LOADABLE_SHARED } from './shared'
+import { register } from './loadableModulesMap';
 
 const STATUS_PENDING = 'PENDING'
 const STATUS_RESOLVED = 'RESOLVED'
@@ -46,8 +47,9 @@ function createLoadable({
   onLoad,
 }) {
   function loadable(loadableConstructor, options = {}) {
-    const ctor = resolveConstructor(loadableConstructor)
     const cache = {}
+    const ctor = resolveConstructor(loadableConstructor)
+    register(ctor);
 
     /**
      * Cachekey represents the component to be loaded
--- a/packages/component/src/loadableModulesMap.js
+++ b/packages/component/src/loadableModulesMap.js
@@ -1,0 +1,36 @@
+let federatedModuleMap = {};
+const localModuleNames = [];
+
+const isFederated = (loadableCtor) => loadableCtor.resolve().indexOf('webpack/container/remote') === 0;
+
+export const register = (loadableCtor) => {
+  const moduleName = loadableCtor.chunkName();
+  if (isFederated(loadableCtor)) {
+    federatedModuleMap[moduleName] = loadableCtor.importAsync;
+  } else {
+    localModuleNames.push(moduleName);
+  }
+};
+
+/**
+ * Calls importAsync for all the modules listed in moduleNames
+ * Resolves after modules are loaded
+ */
+export const preload = (moduleNames) => {
+  const federatedModules = moduleNames
+    .filter((name) => !localModuleNames.includes(name))
+    .map((name) => federatedModuleMap[name]);
+
+  // only need to preload module once
+  federatedModuleMap = {};
+
+  return Promise.all(federatedModules.map((loader) => loader()));
+};
+
+
+/**
+ * Preload all the existing federated modules, to be on the server (we don't have loadable required chunks on SSR)
+ */
+export const preloadAll = () => preload(Object.keys(federatedModuleMap))
+  // repeat if more modules are discovered during preloading
+  .then(Object.keys(federatedModuleMap).length ? preloadAll() : undefined);
--- a/packages/webpack-plugin/src/federatedStatsCollector.js
+++ b/packages/webpack-plugin/src/federatedStatsCollector.js
@@ -1,0 +1,75 @@
+
+const extractSharedModuleOpts = (moduleIdentifier) => {
+  const [
+    _issuerType, _shareScope, _shareKey, semver, _strictVersion, _fullPath, singleton, eager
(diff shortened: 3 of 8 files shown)


packages/component/src/loadableModulesMap.js
```diff
@@ -0,0 +1,36 @@
...
+  } else {
+    localModuleNames.push(moduleName);
+  }
+};
+
+/**
+ * Calls importAsync for all the modules listed in moduleNames
+ * Resolves after modules are loaded
+ */
+export const preload = (moduleNames) => {
```
theKashey, line 19: it cannot be named as `preload` as well as this file cannot be `loadableModuleMap` as long as it's purely focused on the Federation logic.

packages/component/src/loadableModulesMap.js
```diff
@@ -0,0 +1,36 @@
...
+  federatedModuleMap = {};
+
+  return Promise.all(federatedModules.map((loader) => loader()));
+};
+
+
+/**
+ * Preload all the existing federated modules, to be on the server (we don't have loadable required chunks on SSR)
+ */
+export const preloadAll = () => preload(Object.keys(federatedModuleMap))
```
theKashey, line 34: I like this idea. Pretty usable.

packages/component/src/createLoadable.js
```diff
@@ -46,8 +47,9 @@ function createLoadable({
   onLoad,
 }) {
   function loadable(loadableConstructor, options = {}) {
-    const ctor = resolveConstructor(loadableConstructor)
     const cache = {}
+    const ctor = resolveConstructor(loadableConstructor)
+    register(ctor);
```
theKashey, line 52: you are basically making [imported-components](https://github.com/theKashey/react-imported-component) out of loadable.

packages/component/src/loadableModulesMap.js
IMalyugin: Initially named it federatedModulesResolver, but then changed the naming. As for preload, it was initially resolve and resolveAll. But then when i started using named imports it got confusing.
Can change it to federatedResolver.js with:
- register
- resolveNamedFederatedModules
- resolveAllFedratedModules

packages/component/src/loadableModulesMap.js
IMalyugin: Still not sure if it's better than static import for server side. Because static import creates less chunks and makes SSR easier

packages/component/src/createLoadable.js
IMalyugin: Sounds like a good thing, knowing you're the author of imported-components :)

IMalyugin:
Updated PR by adding one more piece of the puzzle - shared eager dependencies flushing to the loadable-stats.
Not yet flushed out on the client, because async boundaries are not processed yet.

packages/webpack-plugin/src/federatedStatsCollector.js
```diff
@@ -0,0 +1,75 @@
...
+  };
+}
+
+const isSharedModule = (moduleIdentifier) => moduleIdentifier.indexOf(`webpack/sharing/`) === 0
+
+const collectSharedModuleConsumptionStats = (modules = []) => {
+  const sharedConsumes = {};
+  const sharedModules = {};
+
+  modules.forEach((module) => {
```
theKashey, line 20: dark magic. I have absolutely no idea what's happening here.

packages/webpack-plugin/src/index.js
```diff
@@ -28,19 +29,36 @@ class LoadablePlugin {
       chunkGroupChildren: true,
       hash: true,
       ids: true,
+      modules: true,
```
theKashey, line 32: this _may_ affect stats size and time required to collect them.
Ideally, extra functionality should be opt-in, not opt-out.

packages/webpack-plugin/src/index.js
```diff
@@ -28,19 +29,36 @@ class LoadablePlugin {
...
-    stats.chunks = [...compilation.chunks].map(chunk => {
-      return {
+    const { sharedConsumes, sharedModules } = collectSharedModuleConsumptionStats(stats.modules);
+    delete stats.modules;
+
+    stats.sharedModules = sharedModules;
+    // stats.sharedConsumes = sharedConsumes;
+
+    stats.chunks = [];
+    [...compilation.chunks].forEach((chunk) => {
```
theKashey, line 47: we are asking `stats` without `chunks`. Why it's not empty?

packages/webpack-plugin/src/federatedStatsCollector.js
IMalyugin: Heh, worth writing a huge comment or a link to github explaination. Thats what took me most of the time, analyzing stats to come up with a way to extract module shared dependencies. They happen to be stored under module.reasons

packages/webpack-plugin/src/index.js
IMalyugin: Not stats size, but time to collect - yes. These modules are later deleted, after I parse them with federatedStatsCollector. Not sure how much time is lost on module collection, but if it's worth it in the end, I believe it is possible to overcomplicate the plugin to test if wmf plugin is also connected. And if it is - enabling module parsing.

packages/webpack-plugin/src/index.js
IMalyugin: That's not originally really my line, I just changed it from map to forEach. Chunks are taken straight from compilation, not from stats (most likely to prevent adding additional stats generation time)

Issue #897, opened 2022-06-01
Passing `hash` and `outputPath` make builds non deterministic


Aghassi:
https://github.com/gregberge/loadable-components/blob/b3a9febff55b02c49ceba602f07fa5128d53fe36/packages/webpack-plugin/src/index.js#L29

We've found that passing `hash` and `outputPath` make the files generated by this plugin non deterministic. Can we either disable it completely by defaulting to `false` or have a preference to override it so that we can keep builds deterministic? This is because the json file generated will have hashes and full file paths in it, which changes per build. We are using bazel and this means that we can't get cache hits even on almost fully deterministic builds.

CC @joeljeske

Aghassi:
I've opened a PR to address this and start discussions with any maintainers. This issue should avoid being marked as stale or won't fix.

Aghassi:
The issue is going to be fixed, it is waiting on review

Aghassi:
@theKashey can you reopen this and please comment on if this or the PR I opened is unreasonable? It sounded like it could be merged but never did and I'm tired of fighting this bot honestly. I'd like to not maintain my internal patch if this is allowed to be merged.

Aghassi:
@fivethreeo Tagging you too if you'd be interested if @theKashey is busy. I see no reason this can't be upstreamed and would like to work to get it in if so.

PR #908, opened 2022-08-14
feat: allow overriding stats for webpack-plugin

Aghassi:
This change allows the end user to pass options
to the stats object to override them for their
given use cases. It will default to the old behavior
if no opts are passed.

Closes https://github.com/gregberge/loadable-components/issues/897

## Summary

Close https://github.com/gregberge/loadable-components/issues/897 in a non breaking way by allowing users to override specific stats opts, but fallback on the defaults if no opts passed.

## Test plan

Open to suggestions in this department.

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -11,8 +11,16 @@ class LoadablePlugin {
     writeToDisk,
     outputAsset = true,
     chunkLoadingGlobal = '__LOADABLE_LOADED_CHUNKS__',
+    loadableStatsFileOptions = {},
   } = {}) {
-    this.opts = { filename, writeToDisk, outputAsset, path, chunkLoadingGlobal }
+    this.opts = {
+      filename,
+      writeToDisk,
+      outputAsset,
+      path,
+      chunkLoadingGlobal,
+      loadableStatsFileOptions,
+    }
 
     // The Webpack compiler instance
     this.compiler = null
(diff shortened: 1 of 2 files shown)


theKashey:
3 points:

1. There is a better way to write desired change in the code
```diff
const stats = compilation.getStats().toJson({
      all: false,
      assets: true,
      cachedAssets: true,
      chunks: false,
      chunkGroups: true,
      chunkGroupChildren: true,
      hash: true,
      ids: true,
      outputPath: true,
      publicPath: true,
+      ...statsOpt
});
```
2. `statsOpt` does not sound like a good name
3. documentation for a feature is completely missing

I would propose start with the last point to display the way you expect this feature to be used, as well as indicate a possible usecase.

Aghassi:
Thanks for the feedback! Let me update this with the above points.

Aghassi:
@theKashey Thanks for the feedback, I've made changes to reflect the points you made. Please let me know if you'd like me to revise anything. Also, point 1 is really neat. I forgot destructuring could work like that!

theKashey:
Great work @Aghassi and absolutely different _picture_ visible to the end customer - another user facing the same problem.

Aghassi:
@theKashey thanks! Is there anything else needed for this to be merged so we can close the issue?

Aghassi:
@theKashey We stumbled over this again internally. Any chance this PR can be merged and released? Is there any more work needed?

theKashey:
Sorry, was a little overloaded in the recent month. Just lost track of stuff to keep my eye on.

Aghassi:
All good! I appreciate the work you do, just didn't want this getting lost! Thank you for your time as always.

Aghassi:
@theKashey can we open the PR and merge it? Should I reopen in a new one?

Issue #718, opened 2021-03-10
React 17 + 18 support


apepper:
Currently `@loadable/component`  only supports react 16 is supported:

https://github.com/gregberge/loadable-components/blob/cba6cab47536e23ded8bac34ee24417e6452344e/package.json#L54-L55

It would be great, if it would also support [React 17](https://reactjs.org/blog/2020/10/20/react-v17.html) and the latest React 18.

apepper:
This is not stale. React 17 is still not supported:

https://github.com/gregberge/loadable-components/blob/73b17fd067579b1c5d38eba00e157964e0930a94/package.json#L54-L55

partoneplay:
I use react@17.0.2 and @loadable/component@5.15.0. They work well.

azangru:
They work, but it shows a dependency conflict when installing packages with npm 7:

![image](https://user-images.githubusercontent.com/6834224/125212068-ca354480-e2a2-11eb-829a-5bb3d523a70f.png)

Please fix this 🙏

IPRIT:
Still relevant

IPRIT:
1️⃣7️⃣

theKashey:
1️⃣8️⃣

mike-parkhomenko:
Hey! Is it possible to let us know approximately date for the next release with React 17,18 support?

Thanks)

abhisharkjangir:
Is it supports react v18? I'm getting below error with 

- "react": "^18.1.0"
- "react-dom": "^18.1.0"
- webpack": "^5.73.0"
- "@loadable/babel-plugin": "^5.13.2",
- "@loadable/component": "^5.15.2",
- "@loadable/server": "^5.15.2",
- "@loadable/webpack-plugin": "^5.15.2",
<img width="1438" alt="image" src="https://user-images.githubusercontent.com/21080263/173178218-db3daec4-0efa-4473-8fa3-638c247e7608.png">

polkovnikov-ph:
@abhisharkjangir No, it doesn't. The library is mostly dead for the last 2 years, except for a few fixed by @theKashey.

azangru:
[...]

PR #911, opened 2022-09-12
#718 server-side-rendering with react 18

vaukalak:
## Summary
th
The issue https://github.com/gregberge/loadable-components/issues/718 mentions, that this library is not usable with react 18. However, this is an update of `server-side-rendering` example within react 18. And it works fine without errors or warnings. From what I can see you should just use react `>=16.3.0`: https://github.com/gregberge/loadable-components/blob/main/packages/component/package.json#L37

## Test plan

```
cd packages/component
yarn
yarn dev
```
open `http://localhost:9000`. Observe that server-side rendered code is properly hydrated without any warnings:
lodable-components-react-18-ssr


theKashey:
The problem with React 18 is more about Suspense. The provided example is not using any "React" orchestration and hence has no issues.

PR #923, opened 2022-11-25
Adds support for server side suspense and streaming

fivethreeo:
## Summary

React 18 now supports  server side suspense and streaming.

This pull updates the babel plugin to transform `lazy`, makes the component suspend server-side and adds two methods to `ChunkExtractor` to incrementally write tags for streaming rendering.

## Test plan

Run the example app at

https://github.com/fivethreeo/loadable-components/tree/feature/server-side-suspense/examples/streaming-server-side-rendering)

### DIFF
--- a/packages/babel-plugin/src/index.js
+++ b/packages/babel-plugin/src/index.js
@@ -41,11 +41,17 @@ const loadablePlugin = api => {
     }
 
     // `loadable.lib()`
-    return (
+    if (
       path.get('callee').isMemberExpression() &&
       path.get('callee.object').isIdentifier({ name: 'loadable' }) &&
       path.get('callee.property').isIdentifier({ name: 'lib' })
-    )
+    ) {
+      return true
+    }
+
+    // `lazy()`
+    return (path.get('callee').isIdentifier({ name: 'lazy' }))
+    
   }
 
   function hasLoadableComment(path) {
--- a/packages/babel-plugin/src/index.js
+++ b/packages/babel-plugin/src/index.js
@@ -49,8 +49,8 @@ const loadablePlugin = api => {
       return true
     }
 
-    // `lazy()`
-    return (path.get('callee').isIdentifier({ name: 'lazy' }))
+    // `lazy(), reactLazy()`
+    return (path.get('callee').isIdentifier({ name: 'lazy' }) || path.get('callee').isIdentifier({ name: 'reactLazy' }))
     
   }
 
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -118,6 +118,18 @@ function createLoadable({
 
       return promise
     }
+    const lazyCachedLoad = props => {
+      const cacheKey = getCacheKey(props)
+      let lazyOrComponent = cache[cacheKey]
+
+      if (!lazyOrComponent) {
+        lazyOrComponent = React.lazy(()=>ctor.importAsync(props))
+        cache[cacheKey] = lazyOrComponent
+
+      }
+
+      return lazyOrComponent
+    }
 
     class InnerLoadable extends React.Component {
       static getDerivedStateFromProps(props, state) {
@@ -151,8 +163,7 @@ function createLoadable({
           if (options.ssr === false) {
             return
           }
-          
-          if (!options.suspense) {
+          if (!options.reactLazy) {
             // We run load function, we assume that it won't fail and that it
             // triggers a synchronous loading of the module
             ctor.requireAsync(props).catch(() => null)
@@ -160,20 +171,16 @@ function createLoadable({
             // So we can require now the module synchronously
             this.loadSync()
 
-            props.__chunkExtractor.addChunk(ctor.chunkName(props))
-
-            return
           }
-
           props.__chunkExtractor.addChunk(ctor.chunkName(props))
-          
+          return
         }
 
         // Client-side with `isReady` method present (SSR probably)
         // If module is already loaded, we use a synchronous loading
         // Only perform this synchronous loading if the component has not
         // been marked with no SSR, else we risk hydration mismatches
-        if (
+        else if (
           options.ssr !== false &&
           // is ready - was loaded in this session
           ((ctor.isReady && ctor.isReady(props)) ||
@@ -186,25 +193,27 @@ function createLoadable({
       }
 
       componentDidMount() {
-        this.mounted = true
+        if (!options.reactLazy) {
+          this.mounted = true
 
-        // retrieve loading promise from a global cache
-        const cachedPromise = this.getCache()
+          // retrieve loading promise from a global cache
+          const cachedPromise = this.getCache()
 
-        // if promise exists, but rejected - clear cache
-        if (cachedPromise && cachedPromise.status === STATUS_REJECTED) {
-          this.setCache()
-        }
+          // if promise exists, but rejected - clear cache
+          if (cachedPromise && cachedPromise.status === STATUS_REJECTED) {
+            this.setCache()
+          }
 
-        // component might be resolved synchronously in the constructor
-        if (this.state.loading) {
-          this.loadAsync()
+          // component might be resolved synchronously in the constructor
+          if (this.state.loading) {
+            this.loadAsync()
+          }
         }
       }
 
       componentDidUpdate(prevProps, prevState) {
         // Component has to be reloaded on cacheKey change
-        if (prevState.cacheKey !== this.state.cacheKey) {
+        if (prevState.cacheKey !== this.state.cacheKey && !options.reactLazy) {
           this.loadAsync()
         }
       }
@@ -318,6 +327,14 @@ function createLoadable({
         } = this.props
         const { error, loading, result } = this.state
 
+        if (options.reactLazy) {
+          return render({
+            fallback,
+            result: lazyCachedLoad(props),
+            options,
+            props: { ...props, ref: forwardedRef },
+          })
+        }
         if (options.suspense) {
           const cachedPromise = this.getCache() || this.loadAsync()
           if (cachedPromise.status === STATUS_PENDING) {
@@ -367,7 +384,10 @@ function createLoadable({
     return loadable(ctor, { ...options, suspense: true })
   }
 
-  return { loadable, lazy }
+  function reactLazy(ctor, options) {
+    return loadable(ctor, { ...options, reactLazy: true })
+  }
+  return { loadable, lazy, reactLazy }
 }
 
 export default createLoadable
--- a/packages/component/src/createLoadable.js
+++ b/packages/component/src/createLoadable.js
@@ -119,6 +119,7 @@ function createLoadable({
       return promise
     }
     const lazyCachedLoad = props => {
+      if (props.__chunkExtractor && options.ssr === false) return null
       const cacheKey = getCacheKey(props)
       let lazyOrComponent = cache[cacheKey]
 
@@ -155,6 +156,11 @@ function createLoadable({
         invariant(
           !props.__chunkExtractor || ctor.requireSync,
           'SSR requires `@loadable/babel-plugin`, please install it',
+        )
+
+        invariant(
+          !options.reactLazy || !props.fallback,
+          'reactLazy cannot be used with fallback, wrap in a Suspense',
         )
 
         // Server-side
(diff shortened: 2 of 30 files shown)


packages/component/src/createLoadable.js
```diff
@@ -151,16 +151,22 @@ function createLoadable({
...
 
-          // We run load function, we assume that it won't fail and that it
-          // triggers a synchronous loading of the module
-          ctor.requireAsync(props).catch(() => null)
+            // So we can require now the module synchronously
+            this.loadSync()
 
-          // So we can require now the module synchronously
-          this.loadSync()
+            props.__chunkExtractor.addChunk(ctor.chunkName(props))
```
theKashey, line 163: You can block here keeping a single `addChunk` after it

packages/component/src/createLoadable.js
```diff
@@ -151,16 +151,22 @@ function createLoadable({
           if (options.ssr === false) {
             return
           }
+          
+          if (!options.suspense) {
```
theKashey, line 155: this is not quite about Suspense, but also about a particular way you render application. For any "old" React API this change will lead to an exception.

We cannot do it that simply. Some external configuration is required.

packages/babel-plugin/src/index.js
```diff
@@ -41,11 +41,17 @@ const loadablePlugin = api => {
...
       path.get('callee').isMemberExpression() &&
       path.get('callee.object').isIdentifier({ name: 'loadable' }) &&
       path.get('callee.property').isIdentifier({ name: 'lib' })
-    )
+    ) {
+      return true
+    }
+
+    // `lazy()`
+    return (path.get('callee').isIdentifier({ name: 'lazy' }))
```
theKashey, line 53: 👍 
Sounds like a bug. It would be great to get this change as a separate PR

packages/server/src/ChunkExtractor.js
```diff
@@ -484,6 +487,32 @@ class ChunkExtractor {
     const assets = this.getPreAssets()
     return assets.map(asset => assetToLinkElement(asset, extraProps))
   }
+
+  getLinkTagsSince(extraProps = {}) {
+    const assets = this.getPreAssets()
+    const linkTags = assets.map(asset => {
+      if (!this.seenChunks.includes(asset.chunk)) {
+        this.seenChunks.push(asset.chunk)
```
theKashey, line 495: Please use `Set` for "seen" stuff. It's more sounds and efficient.

packages/server/src/ChunkExtractor.js
```diff
@@ -484,6 +487,32 @@ class ChunkExtractor {
...
+        return assetToLinkTag(asset, extraProps)
+      }
+      return false
+    })
+    return joinTags(linkTags.filter((tag)=>Boolean(tag)))
+  }
+
+  getScriptTagsSince(extraProps = {}) {
```
theKashey, line 503: Using `since` creates a question `since what`?
It's easier to use a _desctructive_ naming pattern:
- drainScriptTags
- flushScriptTags

packages/server/src/ChunkExtractor.js
```diff
@@ -484,6 +487,32 @@ class ChunkExtractor {
...
+    const requiredScriptTag = this.getRequiredChunksScriptTag(extraProps)
+    const mainAssets = this.getMainAssets('script')
+    const assetsScriptTags = mainAssets.map(asset => {
+      if (!this.seenChunks.includes(asset.chunk)) {
+        this.seenChunks.push(asset.chunk)
+        return assetToScriptTag(asset, extraProps)
+      }
+      return false
+    })
+    return joinTags(assetsScriptTags.filter((tag)=>Boolean(tag)))
```
theKashey, line 513: flipping map and filter makes life easier
```tsx
const assetsScriptTags = mainAssets
  .filter(asset => !this.seenChunks.includes(asset.chunk))
  .map(asset => assetToScriptTag(asset, extraProps));

this.seenChunks.push(...mainAssets.map(({chunk}) => chunk));
```

theKashey:
Great work. But as it changes behavior and extends public API - details has to be discussed.

packages/component/src/createLoadable.js
fivethreeo: I  not sure this is a issue since it is inside if chunkextractor on the server and lazy never worked on the server.

fivethreeo:
The code actually does not suspend server-side. Was a bug that made it work. So loads sync on server.

fivethreeo:
Will redo another way

fivethreeo:
There, should work as advertised with your tips now :)

fivethreeo:
Not quite right yet, ssg false seems not to work for some reason.

packages/babel-plugin/src/index.js
fivethreeo: Need to add some snapshot tests, how do I update snapshots?

theKashey: jest -I, or just delete old ones before running the test

theKashey:
Sound like it’s a good time to extend loadable testing capability in order to handle new scenarios. That always was a little painful to manual test every change as many important scenarios were not covered, and with the upcoming update that would became too much and too easy to break.
> this is a task for another PR and separate stream of work

fivethreeo:
Was thinking I could try setting up some tests with react-testing-library.  Will try to do a pr if that is of interest.

theKashey:
I was thinking about a little more complex webpack based build, but the "complicated" part is why it still not around.
If you can create simpler RTL based solution which can test not "everything", but still "a lot" - that would be perfect.

fivethreeo:
Will make a new issue for testing to move this discussion there https://github.com/gregberge/loadable-components/issues/930

dglozic:
For my education: you added 'flush' variants for LinkTags and ScriptTags, but not for StyleTags. Are they not something that is dynamically discovered during rendering?

Also, to confirm: with these modifications, we can continue to use script preloading in head?

fivethreeo:
> For my education: you added 'flush' variants for LinkTags and ScriptTags, but not for StyleTags. Are they not something that is dynamically discovered during rendering?
> 
> Also, to confirm: with these modifications, we can continue to use script preloading in head?

FlushStyleTags should be added. 

According to https://github.com/reactwg/react-18/discussions/114

You should not use preload.

fivethreeo:
Just needs tests to confirm that the solution works as advertised.

dglozic:
I have seen that, but if so, why are we still collecting link tags? What will those links be for?

fivethreeo:
I think link tags can be used with preload, if I understand correctly css and js will be loaded in order.

dglozic:
Well, as I understand, we should no longer emit link tags for script preloading, and provide script tags as bootstrap scripts in 'renderToPipeableStream' setup.

I am eagerly awaiting a version of Loadable that supports 18 to test these in practice :-).

fivethreeo:
I think, don't use preload / prefetch In webpack. But load css with preload in link tags so script and css load in order. Need to check out what the browser does.

dglozic:
Ah, OK. So compared to the current 16.8+ code, Loadable.Server will return the same styles, links and script tags, only we would need to keep adding them as they are discovered during streaming (if I understood your 'flush' additions).

fivethreeo:
Correct :)

fivethreeo:
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

So maybe fetchpriority "high" and rel "preload stylesheet"

dglozic:
Checking in - any outlook for getting React 18 compatibility along the lines of this PR? We would like to move to 18 but are using Loadable Components and cannot :-(.

fivethreeo:
Loadable needs testing before this can be verified to work. Need to set up e2e testing with cypress or playwright for the example.

fivethreeo:
You could always do a patch package if you want to use it now, but no guarantees.

dglozic:
Thanks, I will wait for the official version I can just NPM install.

theKashey:
narrator voice: 🐌 xmass break is a season of waiting 🍹

fivethreeo:
https://github.com/gregberge/loadable-components/pull/944

dglozic:
@fivethreeo reminder that we need 'Elements' variants of 'Tags' for 'flushX' (flushStyleElements, flushLinkElements, flushScriptElements).

fivethreeo:
@dglozic I am not sure that is needed, because flushing makes little sense when you are rendering instead of streaming html.  When rendering you get all html in one go anyway.

dglozic:
Actually yes, sorry, that was silly. When you flush, it goes directly into the stream and you don't really need Elements. Ignore :-).

fivethreeo:
Been quite sick this week so no time for anything but recuperating. But it is on my radar after I get the e2e testing working since that is needed to verify that this works as it should. So any help on what to test and how to test would help speed it along. I have some ideas ruminating but have not quite figured it out yet. Look at the wip pull for e2e testing.

dglozic:
I like @theKashey suggestion to bump the dependencies to React 18 right away, allowing teams to port to React 18 without making any changes to SSR and the use of Loadable. Once we stabilize, we can try out pre-release versions of loadable-components with support for 'flushXYZ'. It would limit the number of variables to watch for.

dglozic:
So now that dependencies were bumped and I was able to play with React 18.2, I am confused with extending the Writable. Right now, I am not yet doing full streaming: I am using `renderToString` to get the App content and collect the loadable elements (using extractor). Then I am wrapping the app content in JSX for html/head/body and other boilerplate, and streaming the entire thing using `renderToPipeableStream`. That works fine.

Just for kicks, I replaced `response` in pipe with the extended `Writable` as in the example. I am actually not doing anything in it, like this:

```js
	const writable = new Writable({
		write(chunk, encoding, cb) {
			res.write(chunk, encoding, cb);
		},
		final(cb) {
			res.end();
			cb();
		},
		flush() {
			if (typeof res.flush === 'function') {
				res.flush();
			}
		},
		destroy: (err) => {
			res.destroy(err ?? undefined);
		}
	});

	const { pipe } = renderToPipeableStream(rootElement, {
		onShellReady: () => {
			pipe(writable);
		},
		onAllReady: () => {
			writable.write("");
		},
		onError: (err) => {
			_handleError(err, res, callback);
		},
		onShellError: (err) => {
			_handleError(err, res, callback);
		}
	});
```

The above code hangs. If I replace `writable` with `res` it works fine. Any ideas?

fivethreeo:
do you have any suspense or errorbouandries in the app?

dglozic:
I do, yes (an ErrorBoundary React component). No Suspense yet.

dglozic:
As I said, the App renders properly when I call `pipe(res)` in `onShellReady()`, but hangs when I swap `res` with `writable`.

fivethreeo:
Using express?

dglozic:
Yes, and Node LTS 18

dglozic:
Do you think that Express Response has some differences to Node.js Response causing this behaviour?

dglozic:
Never mind, your example also uses Express response for piping.

fivethreeo:
Try taking out the final handler

dglozic:
I figured it out, but it took actually extending the Writable class:

```js
	class MyWritable extends Writable {
		write(chunk, encoding, cb) {
			res.write(chunk, encoding, cb);
		}
	
		writev(chunks, encoding, cb) {
			res.writev(chunks, encoding, cb);
		}
	
		end(chunk, encoding, cb) {
			res.end(chunk, encoding, cb);
		}
	
		flush() {
			if (typeof res.flush === 'function') {
				res.flush();
			}
		}
	
		destroy(err) {
			res.destroy(err ?? undefined);
		}
	}

	const writable = new MyWritable();
```

dglozic:
My speculation is that React calls other Writable methods that were not implemented in the simple constructor, but are when extending like this.

fivethreeo:
Ah, I will update the example in the pull.

dglozic:
I don't know if the example itself is affected or not, I just noticed it myself in my own code. I am confused because I have seen other examples on the web similar to yours, yet I could not get it to work (I suspect something about closing the pipe).

fivethreeo:
Need some jest tests for testing with writeable too. I think a idea is forming on how to test this. Just need to separate all the components into their own app so I can test each kind of loadable component by itself. Then hook into the loadable events I add and do some mocking and rendering with renderToPipeableStream.

thomaswelton:
Hey @fivethreeo thanks for your work so far on this.
Wanted to ask if the stale bot was overzealous, or if it is the case that this feature is now a wontfix and streaming support will not be coming. Thanks.

theKashey:
I reckon we can "combine" this PR and https://github.com/gregberge/loadable-components/pull/976 (in terms of getting rid of the "old" loadable) to get amazing results.
But as any job it requires somebody to work on it. Unfortunately nowadays this is not something I can afford.

Issue #926, opened 2022-11-29
Support Module Federation + React SSR


brunos3d:
## 🐛 Bug Report

There are problems when trying to integrate the loadable with a React application that uses Module Federation, especially if the application performs SSR with federated components.

## Sample to reproduce

I created a repository that exemplifies the issues in two use cases, one with React 16 and the other with React 18. Here is the repo link

https://github.com/brunos3d/loadable-module-federation-react-ssr

### How to run it

To make it work just clone it, navigate with the terminal to one of the example folders, install the dependencies with yarn, and then run `yarn start`, here are the commands to run quickly:

```bash
cd samples/loadable-react-16
yarn install
yarn start
```

## About the sample repo

The project has a [readme file](https://github.com/brunos3d/loadable-module-federation-react-ssr#readme) that explains how to use it, but basically, there are two workspaces in the samples folder, two applications in each workspace.

## Expected behavior

In both samples the expected result is the same: render the `app1` on the server side consuming the Content component that is federated by `app2` application using Module Federation and importing it using loadable. The component must be delivered to the client side statically and must be hydrated so that it can be reactive whenever the user changes the text input.

## Workarounds and hacks

I worked alongside @ScriptedAlchemy so that we could identify a way to make this work and we created some workarounds in the examples to make this work. They are described/listed [here](https://github.com/brunos3d/loadable-module-federation-react-ssr#workarounds-and-hacks), I hope that these applied hacks can be useful to solve the problem natively. The places that have the application of these workarounds are marked with the following comment `// ================ WORKAROUND ================`.

ScriptedAlchemy:
@theKashey So the issue here is that loadable is using require.resolveWeak, we need a way server-side to "flush chunks" in such a manner that the federated `require` call still exists. 

If we do `if(notbrowser) {require('app2/thing')}` and loadable(import(app2/thing)), then loadable works server side since webpack sees the require call as an eager import and will hold startup till those additional chunks are loaded. 

Would it be possible to add a `federated: true` flag to loadable which in turn injects a require() instead of require.resolveWeak on a case-by-case basis?

theKashey:
To clarify - there is no need to change existing code, only add extra explicit `require` into server bundle. I am correct?

ScriptedAlchemy:
To get us unblocked, through the least path of resistance - that's the idea. 
[...]

PR #938, opened 2022-12-14
feat: add moduleFederation option

brunos3d:
## Summary

This PR was created to add support for Module Federation it is originated from this issue: https://github.com/gregberge/loadable-components/issues/926

It aims to add a new option to the [loadabel babel-plugin](https://github.com/gregberge/loadable-components/tree/main/packages/babel-plugin) called `moduleFederation`, it's a boolean option so we just need to set it to `true` to enable the support for Module Federation.

```json
{
  "plugins": [
    ["@loadable/babel-plugin", { "moduleFederation": true }],
    // ...
  ]
  // ...
}
```

This PR also adds some JSDoc typedefs to the plugin, which will help users to understand the plugin better.

## Test plan

To test it you can clone the sample project using [this branch](https://github.com/brunos3d/loadable-module-federation-react-ssr/tree/demo-loadable-fix) as base for the test


packages/babel-plugin/src/index.js
```diff
@@ -20,138 +20,194 @@ const properties = [
...
-  function collectImportCallPaths(startPath) {
-    const imports = []
-    startPath.traverse({
-      Import(importPath) {
-        imports.push(importPath.parentPath)
-      },
-    })
-    return imports
-  }
+/**
```
theKashey, line 23: The golden rule of development is to __never__ refactor and change simultaneously.
Look like this file is not _changed_, however from the git point of view - everything changed.

I also cannot spot any change, but that does not mean there isn't any.

While I do appreciate typings and docs
- [ ] please remove unrelated changes

packages/babel-plugin/src/properties/resolve.js
```diff
@@ -27,11 +47,38 @@ export default function resolveProperty({ types: t, template }) {
...
+  ) => {
+    const properties = path.get('arguments.1.properties')
+
+    const isFederated =
+      moduleFederation ||
+      (Array.isArray(properties) &&
+        properties.some(
+          prop =>
+            prop.isObjectProperty() &&
+            prop.get('key').isIdentifier({ name: 'federated' }) &&
```
theKashey, line 67: Not sure how this work and when it should be used.

Please keep in mind that in no circumstances we can use `federated` template on the client-side bundle - that will effectively disable code splitting.
Right now, there is a big probability that it will not happen.

theKashey:
Per my understanding, the essence of this change is to disable code splitting for federation, not to "support" federation keeping the current constraints.

The question is how to ensure this configuration is used properly - only for server-side bundle.

packages/babel-plugin/src/index.js
brunos3d: Cool, thanks for the tip! Done :rocket:

packages/babel-plugin/src/properties/resolve.js
brunos3d: Yeah, make sense, the idea now is to use `moduleFederation: true` option only on server-side compiler

ScriptedAlchemy:
woud we set a DefinePlugin like 'process.isBrowser' and if its false webpack could remove the dead code? Not sure if that would work or not

brunos3d:
@theKashey yeah, your right about using it only on server-side, I will work to make it work as expected, don't have ideas on how we can make it

brunos3d:
@theKashey just removed every unrelated change and forced push it to avoid overwriting the git history, I also removed the federated option, now we need to figure out how to ensure that it uses `require(ID)` only on server-side

brunos3d:
@theKashey just added the suggested warnings. As I'm not too used to babel/webpack plugins I will ask you for code suggestions to solve this issue without workarounds :smile:

packages/babel-plugin/src/properties/resolve.js
```diff
@@ -42,8 +42,27 @@ export default function resolveProperty(
...
+          console.warn(
+            'The "moduleFederation: true" option will disable code splitting on the client-side',
+          )
+        }
+      } else {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'It\'s recommended to use "isBrowser" global variable to detect the environment\n' +
+            'Try to add "webpack.DefinePlugin({ \'process.isBrowser\': true })" to your webpack config',
+        )
```
bwhitty, line 61: Is there something internal to a "client side" Webpack `target` that can be read automatically? For example, as a developer if I set `target: "web"` (the default) or another client-targetted value, then this plugin should probably just error out here (or have a "force" option to override).

Or does Webpack not provide a standard way to expose "is browser" and every webpack tool has to invent their own? (like [here](https://github.com/visgl/luma.gl/blob/v6.0.0/src/utils/is-browser.js))

theKashey: One can use `package.json:browser` field to detect browser target - https://github.com/theKashey/detect-node/blob/master/package.json#L6-L10
But this is a run-time operation.

brunos3d: These solutions only work during runtime, @theKashey do you know if webpack provides some context to babel during the compilation?

theKashey: According to [babel-loader source](https://github.com/babel/babel-loader/blob/main/src/index.js#L54) - no, it does not.

bwhitty: Hm, this begs the question then: can this be an argument to `LoadablePlugin` webpack plugin instead, and then read somehow by the Babel plugin? Is that possible? 

That would alleviate most concerns of misuse, as it could be documented to just use this option for your client-side Webpack build, which is something already natural to loadable components. 

With a Babel plugin option, users have to implement some logic to modify their Babel options dynamically based on whether it's being used by the Webpack server build or client build, which while not weird is an additional configuration workflow added to use loadable, and they also have to implement some `DefinePlugin` thing, which again is another workflow change.

No big deal, just wondering if that could simplify things

theKashey: 🤔 BINGO!
`LoadablePlugin` can automatically call `DefinePlugin` with the variables required. As webpack plugin is expected to be used only for client side it can secure dead code elimination (in production only) and other checks to guide proper configuration.

brunos3d: Awesome @theKashey! I will work on that here! :rocket:

brunos3d: @theKashey added DefinePlugin to LoadablePlugin and updated the warning messages

packages/babel-plugin/src/properties/resolve.js
```diff
@@ -27,11 +41,35 @@ export default function resolveProperty({ types: t, template }) {
...
   }
 
-  return ({ callPath, funcPath }) =>
-    t.objectMethod(
+  return ({ callPath, funcPath }) => {
+    const { isBrowser } = process
+
+    if (moduleFederation) {
+      if ('isBrowser' in process) {
+        if (isBrowser === true) {
```
bwhitty: could be simplified
```suggestion
    if (moduleFederation && process?.isBrowser === true) {
``

File: packages/webpack-plugin/src/index.js [hunk 5]
```diff
@@ -101,6 +101,19 @@ class LoadablePlugin {
   apply(compiler) {
     this.compiler = compiler
 
+    const { webpack } = compiler
+
+    const defs = {
+      'process.isBrowser':
+        compiler.options.target === 'web' ||
+        compiler.options.target === undefined,
+    }
```
bwhitty, line 110: To go with my suggestion of adding a `serverSideModuleFederation` argument to the webpack plugin, if we do then this logic could throw an error for misconfiguration:

```suggestion
   if (this.opts.serverSideModuleFederation === true &&
     compiler.options.target === 'web' ||
     compiler.options.target === undefined
   ) {
     throw new Error(`Enabling server-side module federation support for a client-side Webpack target (${compiler.options.target}) is unnecessary and will effectively disable all code-splitting.`
   }
```

packages/webpack-plugin/src/index.js
```diff
@@ -101,6 +101,19 @@ class LoadablePlugin {
...
+
+    // eslint-disable-next-line global-require
+    new ((webpack && webpack.DefinePlugin) || require('webpack').DefinePlugin)(
+      defs,
+    ).apply(compiler)
```
bwhitty, line 115: with my suggestion this would become

```suggestion
  if (this.opts.serverSideModuleFederation) {
    // eslint-disable-next-line global-require
     new ((webpack && webpack.DefinePlugin) || require('webpack').DefinePlugin)({
       serverSideModuleFederation: true
     }).apply(compiler)
  }
```

bwhitty (commented):
Now that there's a webpack plugin integration, what's the purpose of the Babel plugin argument? Is there a use case for a standalone Babel build for Loadable without Webpack? If not, then it seems the Webpack plugin can be the singular configuration point.

I'm thinking that this federation config can now be set up in one place: the server-side Webpack configuration's `LoadablePlugin`. Like so:

```
// webpack.config.server.js

plugins: [
  new LoadablePlugin({
    serverSideModuleFederation: true
  }),
]
```

Then internally the plugin would set `process.serverSideModuleFederation = true`, and then `@loadable/babel-plugin` can just read that. There's no need to set up a separate config on the Babel plugin, as it's configured via Webpack.

packages/webpack-plugin/src/index.js
brunos3d: Nice job @bwhitty, I will test it here!! :rocket:

brunos3d: @theKashey I am not being able to get the value of `serverSideModuleFederation` passed to DefinePlugin during the execution of the resolveProperty function, do you have some idea on how can I access this value?
I already tried with:
- process.serverSideModuleFederation
- global.serverSideModuleFederation
- only serverSideModuleFederation

brunos3d: @ScriptedAlchemy I got these snippets from [NodeFederationPlugin](https://github.com/module-federation/universe/blob/1ca2c86a51167cc792d33342968a411dad879275/packages/node/src/plugins/NodeFederationPlugin.ts#L161-L172) do you have some idea why I can't get the serverSideModuleFederation during babel compilation execution?

packages/webpack-plugin/src/index.js
```diff
@@ -101,6 +109,26 @@ class LoadablePlugin {
   apply(compiler) {
     this.compiler = compiler
 
+    const { webpack } = compiler
+
+    if (
+      (this.opts.serverSideModuleFederation === true &&
+        compiler.options.target === 'web') ||
+      compiler.options.target === undefined
```
bwhitty, line 117: these parens make this logic wrong I think? should be
```suggestion
      this.opts.serverSideModuleFederation === true &&
         (compiler.options.target === 'web' || compiler.options.target === undefined)
```

brunos3d: Yeah, it was not correct, just fixed, but it was not the problem

brunos3d:
Hey, folks as we didn't find a proper way to pass the `serverSideModuleFederation` option from the LoadablePlugin (webpack) to babel, I just decide to inject the value to process using directly, as it will be used only during the compilation time, I can't see any problems on that solution. Excited to see your suggestions about it
cc: @theKashey @bwhitty @ScriptedAlchemy

ScriptedAlchemy:
@theKashey any input here - would be great to move this forwards if we can

packages/webpack-plugin/src/index.js
```diff
@@ -101,6 +109,23 @@ class LoadablePlugin {
...
+      (compiler.options.target === 'web' ||
+        compiler.options.target === undefined)
+    ) {
+      throw new Error(
+        `Enabling server-side module federation support for a client-side Webpack target (${compiler.options.target}) is unnecessary and will effectively disable all code-splitting.`,
+      )
+    }
+
+    if (this.opts.serverSideModuleFederation) {
+      Object.defineProperty(process, 'serverSideModuleFederation', {
```
theKashey, line 123: 👍 that's a fancy way to "teleport" settings.
However, the particular way you did it will break some _assumptions_ of node creating a hard to understand "gap" for a few (rare) users of thread-loader or similar stuff.

Ease to fix. Just store what is required in [process.env](https://github.com/JaneaSystems/node/blob/2df82581162337f9e402d55f2d2d87585cb81dea/lib/child_process.js#L518) - a special place _designed_ for this.

theKashey (approved):

theKashey:
Before we processed with the merge, I want to ask one question - _Where_ is the best place to handle this stuff?
Having the problem of "federation gap", ie one's inability to simply require yet undefined stuff, there are two options:
- make a sync require in the loadable (this PR). That would solve problem only for loadable and in a very specific way.
- preload all known components in the loadable (not which are not currently tracked)
- preload all federated sources via `ModuleFederationPlugin`. Ie go via all known remotes and import them before starting a server

fivethreeo:
This does module federation for all modules. Should we do it as a option on loadable call?

theKashey:
So that is the question: if one technology should adapt to support another technology - could there be a third technology to change as well?
For example, @fivethreeo is working on React 18 update with a "proper suspense" on the server side - https://github.com/gregberge/loadable-components/pull/923 - which technically can allow us to use plain `import` without `require.resolve`

ScriptedAlchemy:
Proper suspense would work for everything moving forwards, but we would still need a solve for async-node target / lower react versions :P

brunos3d:
@theKashey @ScriptedAlchemy An issue we're facing right now is when we try to consume a container from some remote using loadable and the remote is down, MF tries to generate a fake remote and a fake container, but loadable throws multiple errors saying that the provided module is not a react component (cause it's an empty object), an a bunch of other errors.

image

image

image

image

By now I'm using the following workaround for local linked and patched @loadable/component
```js
// on InnerLoadable component at the top of the "render" function

if (result && typeof result === 'object' && Object.keys(result).length === 0) {
  console.log('probably the fake container')
  result = `<span>failed to load remote container</span>`
}
// ...
```
It checks if the result (the loaded/resolved module) is not undefined and verifies if its an empty object (the fake container generated by MF) then it replaces it with the warning message

theKashey:
There is a `resolveComponent` option you can provide to loadable, and the one capable to check imported object and throw and error (to emulate load failure, as it is a failure). Wondering if can call another check just after it and throw an error if component was resolved to nothing. That should trigger "normal" error branches (`fallback` or `ErrorBoundary`), not break `createElement`.
> A little different issue, but quite related

ScriptedAlchemy:
@brunos3d i might be able to help with the fake remote which returns null. We could modify node federation or allow it to throw a rejection. Generally I avoid that since throwing errors in sync imported components can cause webpack runtime to crash

ranshamay:
great work!
@gregberge  any plan to merge it soon? :)

brunos3d:
@ranshamay Probably we will not merge this PR, but I'm working on some stuff:

- An article (WIP)
- These [loadable adapters](https://www.npmjs.com/package/@mf/loadable-adapters)
- This [repo with samples ](https://github.com/brunos3d/react-mf-loadable-ssr)on how to use @mf/loadable-adapters

I will post the updates here soon as I finish everything, but its also good to keep an eye on https://valor-software.com/articles too 🚀

oguzhanaslan:
@brunos3d I saw this example that you prepared: [https://github.com/module-federation/module-federation-examples/tree/88c252fbc3fdc3262e0d8e3d03a945276bd0a165/loadable-react-16](url)

In my server-side micro frontend application, I will use module federation to combine shared libraries (react, react-dom, axios). Is it problematic to use this? I am eagerly waiting for your answer.

PR #944, opened 2023-01-10
WIP: Cypress e2e example test

fivethreeo:
## Test plan
Make sure you built loadable first

Run `rm -rf node_modules yarn.lock`,  `yarn` in `server-side-rendering` example

Run `yarn build:test`, `yarn test:cypress`, `yarn test:jest` in `server-side-rendering` example


packages/component/src/createLoadable.js
```diff
@@ -258,6 +260,8 @@ function createLoadable({
           const result = resolve(loadedModule, this.props, Loadable)
           this.state.result = result
           this.state.loading = false
+          console.log(
```
theKashey, line 263: So I was thinking about making loadable a little more _observable_ and letting someone from the outside collect metric from it:
- when and what it loads
- how much time it takes
- so on

From "this file" point of view it's the same console.log - a call to some function with some "event". And then you can _expect_ this call to happen in tests enabling you to create them.

Without such _observability_ you cannot really create "platform" tests, only something very smoke application-level one. Well, testing the outcome might sounds like a good idea, but at the library level "details matter"

Thoughts?

fivethreeo: Was thinking of having a dev build so all test code is removed in prod. But observability would be good.

theKashey: Dev build something like something achievable with fewer efforts.

fivethreeo: Added a commit now, something like that?

packages/component/src/loadableEvents.js
```diff
@@ -0,0 +1,9 @@
+import EventEmitter from 'events'
```
theKashey, line 1: that is not the best package for a clientside project

Ideally make this one configurable from the outside:
- expose (to a client) a command `configureLoadableEventsController` which will accept a simple function accepting an `event`
- expose (internally) command `emitEvent` calling provided callback, or doing nothing if it was not set
- replace `loadableEvents.emit(` by `emitEvent` and worry less about complicated types or pulling in some node-related packages requiring a polyfill for browser space.

That is a general rule of interface segregation principle - do not ask (and do not provide) more than one should. Loadable needs an ability to emit a single event, let's all it a 🍌 , but you are giving it a full featured event managers - 🦍 and the rest of the 🌴 
> This is a reference to https://www.johndcook.com/blog/2011/07/19/you-wanted-banana/

fivethreeo:
More like this?

packages/component/src/createLoadable.js
```diff
@@ -5,6 +5,7 @@ import hoistNonReactStatics from 'hoist-non-react-statics'
 import { invariant } from './util'
 import Context from './Context'
 import { LOADABLE_SHARED } from './shared'
+import * as loadableEvents from './loadableEvents'
```
theKashey, line 8: this moment need only "emit" part, import what you need.

packages/component/src/index.js
```diff
@@ -10,7 +11,7 @@ const { lazy } = loadableExports
 lazy.lib = libraryExports.lazy
 
 export default loadable
-export { lazy }
+export { lazy, loadableEvents }
```
theKashey, line 14: You are exposing `emit` to the end user, and this is something they should not use.
I like `loadableEvents.registerCallback` signature, but exposing internals is not great.

theKashey:
> More like this?

You are reading my mind

fivethreeo:
Added jest test but event callback mock seems not to be registered

fivethreeo:
Jest tests are working

examples/server-side-rendering/cypress/tests/spec.cy.js
```diff
@@ -0,0 +1,37 @@
...
+    })
+    cy.window().then((win) => {
+      cy.spy(win, 'loadableEvents').as('spyloadableEvents')
+    })
+  
+
+    cy.get('@spyloadableEvents')
+      .should('be.calledWith', Cypress.sinon.match({ type: 'loadSync', chunkName: 'letters-A' }))
+      .should('be.calledWith', Cypress.sinon.match({ type: 'loadSync', chunkName: 'letters-F' }));
+  })
```
theKashey, line 24: if you abstract this operation to a local helper you can make it shorter and get more "clear" signal as a result.

Good testing is all about removing _noise_ from the code in order to reveal intentions

theKashey (approved):

examples/server-side-rendering/cypress/tests/spec.cy.js
fivethreeo: Not quite sure what that would look like. Got any pointers or ideas? 😀

theKashey: The general concept is perfectly described here - https://www.cypress.io/blog/2019/01/03/stop-using-page-objects-and-start-using-app-actions/ and https://glebbahmutov.com/blog/readable-tests/

About readability and _extraction of helpers_ to maximise signal from tests.
Cypress usually proposes custom commands, but I prefer plain functions as they are less "magical".

long story short
```tsx
// extract helper
const shouldObserve = (event) => {
   cy.get('@spyloadableEvents')
       .should('be.calledWith', Cypress.sinon.match(event))
}

// use helper

shouldObserve({ type: 'loadSync', chunkName: 'letters-A' });
shouldObserve({ type: 'loadSync', chunkName: 'letters-F' });
```

So there is almost no difference, but you are removing unnecessary details from the code and can focus on what matters more.

Issue #950, opened 2023-02-02
Add importEntrypoint to CunkExtractor and don’t load stats with require.


fivethreeo:
## 🚀 Feature Proposal

Add importEntrypoint to CunkExtractor and don’t load stats with require.

## Motivation

For module federation and testing using import to get the entrypoint would be preferable to using require.

## Example

```js
const App = await extractor.importEntrypiont(‘test entrypoint')
// stats should be read with fs so there are no requires if we use es modules
renderToPipeableStream(App, res)
```

theKashey:
is it technically different?
This will blow up a little with transpilation/cjs so what about moving _responsibility_ to the end user
```tsx
const App = extractor.importEntrypiont(await import('test entrypoint'))
//                                     ^ everything under user control
```

fivethreeo:
Yeah that seems better.

fivethreeo:
```js
const App = await extractor.importEntrypiont(‘test entrypoint', (entry) => import(entry)))
```

Maybe?

theKashey:
That will not be statically analyzeable. Not sure if there's a need, it's not analyzable right now, but why you might need such indirection?
Lets apply occam's razor here and cut what we don't have a good reason for.

dglozic:
Also a typo in the example (piont).

fivethreeo:
```js
const App = await import(extractor.getEntrypointPath(‘testentrypoint'))
```
Is better perhaps.

fivethreeo:
requireEntrypoint seems to require all script assets up front, so should it import all too?

fivethreeo:
In the pull i added you can now do

```js
await Promise.all(extractor.getAllScriptAssetsPaths().map(asset=>import(asset)))
const App = await import(extractor.getEntrypointPath('testentrypoint'))
```

fivethreeo:
Fixed

PR #951, opened 2023-02-02
Added getEntrypointPath to ChunkExtractor and reading json from fs

fivethreeo:
#950

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -383,12 +383,11 @@ class ChunkExtractor {
   // Utilities
 
   requireEntrypoint(entrypoint) {
-    const entrypointPath =this.getEntrypointPath(this.entrypoint)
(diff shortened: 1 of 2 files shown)


theKashey (approved):
Very good move 👍

packages/server/src/ChunkExtractor.js
```diff
@@ -205,7 +205,7 @@ class ChunkExtractor {
     inputFileSystem = fs,
```
theKashey, line 205: 🤔 🤔 🤔 this symbol is __undefined__

fivethreeo: Undefined? fs is imported. Errors?

theKashey: 🤦‍♂️ imported above the fold of github review

theKashey:
👍 👍 👍

PR #966, opened 2023-06-28
Allow additional identifiers to be passed in to the babel plugin

edkimmel:
## Summary

I would like to wrap `loadable` in my own higher order component functions.  Loadable by default only injects a webpack chunk name  for `loadable` and `lazy`, but can readily accept additional identifiers.

This PR allows us to rewrite

```
// Loadable
const LoadableHelpPage = loadable(() => import('./HelpPage'), {
  resolveComponent: components => components.HelpPage,
})
// Our custom HOC
const HelpPage = PageHOC(LoadableHelpPage, {...additional page props })
```
to
```
const HelpPage = LoadablePageHOC(() => import('./HelpPage'), 'HelpPage', { ...additional page props })
```

We actually have a separate HOC for modals, so we would not be able to simply rename LoadablePageHOC to loadable for the sake of the existing babel plugin.  With the changes, we can specify `additionalIdentifiers: [ 'LoadablePageHOC', 'LoadableModalHOC']` to transform both.

## Test plan

New tests have been added that showcase the custom identifier being transformed or not, depending on babel options.

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -232,6 +232,64 @@ exports[`plugin Magic comment should transpile shortand properties 1`] = `
 };"
 `;
 
+exports[`plugin additionalSpecifiers Should match on custom specifiers 1`] = `
+"LoadableHOC({
+  resolved: {},
+
+  chunkName() {
+    return \`ModA\`.replace(/[^a-zA-Z0-9_!§$()=\\\\-^°]+/g, \\"-\\");
+  },
+
+  isReady(props) {
+    const key = this.resolve(props);
+
+    if (this.resolved[key] !== true) {
+      return false;
+    }
+
+    if (typeof __webpack_modules__ !== 'undefined') {
+      return !!__webpack_modules__[key];
+    }
+
+    return false;
+  },
+
+  importAsync: () => import(
+  /* webpackChunkName: \\"ModA\\" */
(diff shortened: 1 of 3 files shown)


theKashey:
Got your idea, the request seems quite familiar 👍 and look like the currently proposed way to implement own loadable - [loadable-detection](https://loadable-components.com/docs/babel-plugin/#loadable-detection) - is not working. At least looking at the code.

In short - a right move, but let's take one more step and perfect this solution so it would not backfire in a couple of years as I don't feel that replacing _random functions_ is a very safe pattern. 
What about changing the configuration a little bit to replicate the current logic of "autodetecting" loadable imports, but for a custom function?

Ie
`additionalIdentifiers: [ 'LoadablePageHOC', 'LoadableModalHOC']`
⬇️ 
```tsx
signatures: [ 
  {name:'LoadablePageHOC', from:'@myapp/loadable'},
  {name:'LoadableModalHOC', from:'@myapp/loadable'},
  {name:'default', from:'@loadable/components'}, // still support default way
]
```
One can also refactor the current logic "looking" for `default` import of loadable and named import of lazy to utilize the same configuration, with your ability to totally override default configuration

edkimmel:
@theKashey Thank you for the review.

Your suggestions are great, I should hopefully have some time this week to make the changes.

theKashey:
Amazing result 👍 
Let's get it merged and I will update the documentation shortly after.

theKashey:
Released as `5.16.0`

PR #972, opened 2023-07-17
fix: correct babel plugin default signature to allow any source of 'loadable'

theKashey:
The old behaviour was allowing `loadable` to be of any source, even a local variable. Signatures tightened this logic and broke some user code.

Fix: default to `loadable` name unless signatures are overridden.

### DIFF
--- a/packages/babel-plugin/src/index.js
+++ b/packages/babel-plugin/src/index.js
@@ -20,12 +20,11 @@ const properties = [
 
 const LOADABLE_COMMENT = '#__LOADABLE__'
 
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/index.js
```diff
@@ -124,7 +123,8 @@ const loadablePlugin = declare((api, {
       Program: {
         enter(programPath) {
           let lazyImportSpecifier = false
-          const loadableSpecifiers = []
+          // default to "loadable" detection. Remove defaults is signatures are configured
```
theKashey, line 126:
```suggestion
          // default to "loadable" detection. Remove defaults if signatures are configured
```

PR #976, opened 2023-08-21
Add api exposing a hook-based way to load a module

edkimmel:
## Summary

Adds loadable.hook and lazy.hook as additional ways to load a dynamic module via @loadable

A common use case is to use some large third party library as part of a hook in first party code.  By exposing a hook-based API in @loadable, it becomes trivial to dynamically load that library module.

```
const useThirdPartyAuthSdk = loadable.hook(() => 'some-large-sdk')
...

const useUserSignIn = () => {
  // 'some-large-sdk' is not loaded until this hook is ran
  const thirdPartyAuthSdk = useThirdPartyAuthSdk()
  const signin = React.useCallback(creds => {
    if (!thirdPartyAuthSdk) throw new Error('Auth SDK used before ready')
    thirdPartyAuthSdk.initiateSignin(creds)
  }, [thirdPartyAuthSdk])
  return {
    ready: !!thirdPartyAuthSdk,
    signin
  }
}
```

## Test plan

`loadableHook.test.js` has been created mirroring `loadable.test.js` where applicable.
An example for the hook api was created in the repository.

### DIFF
--- a/packages/babel-plugin/src/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin/src/__snapshots__/index.test.js.snap
@@ -881,6 +881,177 @@ exports[`plugin custom signatures should not match on undeclared specifiers 1`] 
 myLoadable(() => import(\`./ModA\`));"
 `;
 
+exports[`plugin lazy.hook should be transpiled too 1`] = `
+"import { lazy } from '@loadable/component';
+lazy.hook({
+  resolved: {},
+
+  chunkName() {
+    return \\"moment\\";
+  },
+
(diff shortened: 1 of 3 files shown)


edkimmel:
TODO: Update the babel plugin to catch the new pattern.

theKashey:
Amazing job and quite interesting implementation. Big kudos for comprehensive tests.

However, let me be clear - I am going to add this PR to the list of all other things one way or another assigned to me and I cannot provide you any ETA for actual resolution. My presence in GitHub lands is quite limited nowadays.

Issue #999, opened 2024-02-08
SSR is broken for component v5.16.2+


DonGissel:
## 🐛 Bug Report

Upgrading from `@loadable/component` `v5.15.3` (latest `5.15.*` in npm) to `v5.16.2+` quietly breaks SSR using React 17 and Webpack 5.90.

## To Reproduce

Something has broken between `@loadable/component` `v5.15.3` and `v5.16.2` (there is no `5.16.0` or `5.16.1` in npm, so `5.15.3` -> `5.16.2/3` is the only upgrade path available). SSR yields absolutely no output for the components, and there are no collected chunks in the output (rehydrating to interactive seems to fix the problem, but that's not really enough).

Using `@loadable/component` `5.15.3` together with `@loadable/server` `5.15.3` _and_ `5.16.3` works perfectly, so the problem is not on the server part. No configuration has changed, but somehow, ~Palpatine returned~ it has just stopped working. If I downgrade `@loadable/component` to `5.15.3`, everything works again.

## Expected behavior

I'd expect it to... you know, render the components on the server. 😅 I looked into the changelog for the repository, and saw the main difference for the client was an added `loadable` named export, so one should not have to use the default exports, but in the versions available on npm, I can't see `loadable` being exported (and the compiler explodes when I try accessing it). The types available through `@types/loadable__component` and `@types/loadable__server` have not been updated recently either, so I think the default export is still the way to go when using the regular `import` pattern (ES6/ESM).

## Link to repl or repo (highly encouraged)

Sadly, it's a proprietary client repo, but since everything else is unchanged, I hope/fear it's reproducible anywhere? Or it might be something in this unique setup that's broken, but it's weird that it'd only break in the relatively short jump from 5.15.3 to 5.16.2?

> Issues without a reproduction link are likely to stall.

I'll take my chances, but thanks.

## System info

```bash
## System:
 - OS: Windows 11 10.0.22631
 - CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
 - Memory: 14.02 GB / 31.76 GB
## Binaries:
 - Node: 20.11.0 - C:\Program Files\nodejs\node.EXE
 - npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
## npmPackages:
 - @loadable/babel-plugin: ^5.16.1 => 5.16.1
 - @loadable/component: 5.16.3 => 5.16.3
 - @loadable/server: 5.16.2 => 5.16.2
[...]

PR #1003, opened 2024-04-08
Migrate server to ESM

7rulnik:
## Summary

Fixes https://github.com/gregberge/loadable-components/issues/999

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,10 +5,23 @@ function getTargets() {
   return undefined
 }
 
+function getModules() {
+  if (process.env.MODULE_TARGET === 'cjs') {
+    return 'cjs'
+  }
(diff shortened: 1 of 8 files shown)


packages/component/package.json
```diff
@@ -44,7 +44,7 @@
     "react": "^16.3.0 || ^17.0.0 || ^18.0.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.7.7",
+    "@babel/runtime": "^7.12.18",
```
7rulnik, line 47: 7.12 is the first version that introduces `exports` field in package.json. We need it for ESM

theKashey (approved):
That's the best solution we have so far.
Let's look forward to move everything to a single build platform. But not today

Issue #1005, opened 2024-04-20
Webpack doesn't like @loadable/server 5.16.4


Judahmeek:
```
webpack  | ERROR in ./node_modules/@loadable/server/dist/esm/loadable-server.esm.mjs 6:0-31
webpack  | Module not found: Error: Can't resolve 'lodash/uniq' in '/Users/judahmeek/apps/datacenters/node_modules/@loadable/server/dist/esm'
webpack  | Did you mean 'uniq.js'?
webpack  | BREAKING CHANGE: The request 'lodash/uniq' failed to resolve only because it was resolved as fully specified
webpack  | (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
webpack  | The extension in the request is mandatory for it to be fully specified.
webpack  | Add the extension to the request.
```

### Results of `npx envinfo --system --binaries --npmPackages webpack,@loadable/component,@loadable/server,@loadable/webpack-plugin,@loadable/babel-plugin --markdown --clipboard`:

```bash
## System:
 - OS: macOS 14.4.1
 - CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
 - Memory: 2.19 GB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
 - Yarn: 4.1.1 - ~/.nvm/versions/node/v18.16.0/bin/yarn
 - npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
## npmPackages:
 - @loadable/babel-plugin: ^5.16.1 => 5.16.1 
 - @loadable/component: ^5.16.4 => 5.16.4 
 - @loadable/server: ^5.16.4 => 5.16.4 
 - @loadable/webpack-plugin: ^5.15.2 => 5.15.2 
 - webpack: ^5.90.3 => 5.90.3 
```

I'll get a reproduction repo up ASAP.

theKashey:
😭 that should not be required

theKashey:
Thank you for the timely report. The root cause is related only to lodash as it a little specific package without any entry points defined. Easier to fix than verify the change

wojtekmaj:
[...]

PR #1006, opened 2024-04-20
Bring back extensions in server package

wojtekmaj:
## Summary

Brings back extensions removed in 114cea2b61ef644d3b794a9b5f48df69f0f5400b, which caused issue #1005.

## Test plan

Warning! The infra (particularly Webpack 4) did not let me even install the repository, so this is a "blind fix".

Testing should be fairly simple though if someone is able to:

- Build the package
- Open ESM file from dist directory using node (e.g. `node ./dist/esm/loadable-server.esm.mjs`)
- Observe if it doesn't crash upon module resolution

### DIFF
--- a/packages/server/src/ChunkExtractor.js
+++ b/packages/server/src/ChunkExtractor.js
@@ -1,13 +1,13 @@
 /* eslint-disable react/no-danger */
 import path from 'path'
 import fs from 'fs'
-import uniq from 'lodash/uniq'
-import uniqBy from 'lodash/uniqBy'
-import flatMap from 'lodash/flatMap'
+import uniq from 'lodash/uniq.js'
+import uniqBy from 'lodash/uniqBy.js'
+import flatMap from 'lodash/flatMap.js'
 import React from 'react'
-import { invariant, getRequiredChunkKey } from './sharedInternals'
-import ChunkExtractorManager from './ChunkExtractorManager'
-import { smartRequire, joinURLPath, readJsonFileSync } from './util'
+import { invariant, getRequiredChunkKey } from './sharedInternals.js'
+import ChunkExtractorManager from './ChunkExtractorManager.js'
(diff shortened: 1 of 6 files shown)


packages/server/src/ChunkExtractor.js
```diff
@@ -1,13 +1,13 @@
 /* eslint-disable react/no-danger */
 import path from 'path'
 import fs from 'fs'
-import uniq from 'lodash/uniq'
-import uniqBy from 'lodash/uniqBy'
-import flatMap from 'lodash/flatMap'
+import uniq from 'lodash/uniq.js'
+import uniqBy from 'lodash/uniqBy.js'
+import flatMap from 'lodash/flatMap.js'
```
theKashey, line 6: only external imports need a full specification as everything internal ends up bundled in one file.

The only real problem here - I could not reproduce the problem to verify the fix is "enough"

wojtekmaj: Too many extensions are far less harmful than too few :)

theKashey:
> Open ESM file from dist directory using node (e.g. node ./dist/esm/loadable-server.esm.mjs)

```
import { __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED } from '@loadable/component';
import _extends from '@babel/runtime/helpers/esm/extends';
import path from 'path';
```

> Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/akorzunov/dev/github/loaders/loadable-components/node_modules/@babel/runtime/helpers/esm/extends' imported from /Users/akorzunov/dev/github/loaders/loadable-components/packages/server/dist/esm/loadable-server.esm.mjs
Did you mean to import @babel/runtime/helpers/esm/extends.js?

Babel reconfiguration is required...

Goodup302:
Hello @theKashey or @all, I don't know if you have this permission, but is it possible to publish version 5.16.5 on npm as soon as possible?

PR #1013, opened 2024-09-06
Move `initialChunks` store variable to window global variable

ahce:
## Summary

Supports calling `loadableReady` once, when there are multiple hydration entry points.

### Example:

- main entry:
```js
initSomething()

loadableReady(() => {
  window.dispatchEvent(new CustomEvent('loadable:ready'))
})
```
- app1 entry:
```jsx
window.addEventListener('loadable:ready', () => {
  hydrateRoot(document.getElementById('app-1'), <App1 />)
})
```
- app2 entry:
```jsx
window.addEventListener('loadable:ready', () => {
  hydrateRoot(document.getElementById('app-2'), <App2 />)
})
```

### Note

To make this with the current version, the `namespace` property must be implemented for each entry with the `loadableReady` call, also the required chunks are duplicated for each JSON script when the different apps use the same component.

## Test plan

No changes were applied that modify the current behavior.

### DIFF
--- a/packages/component/.size-snapshot.json
+++ b/packages/component/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/cjs/loadable.cjs.js": {
-    "bundled": 16900,
-    "minified": 7226,
-    "gzipped": 2553
+    "bundled": 17103,
+    "minified": 7356,
+    "gzipped": 2583
   },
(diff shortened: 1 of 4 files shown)


ahce:
Hi @theKashey , can you review this please? 🙏

PR #1035, opened 2026-04-27
fix: coerce chunk ids to string before comparison

panbenson:
## Summary
Upon bumping to Rspack V2, projects using loadable become unresponsive/stuck in loading. Rspack V2 switched to numeric chunk ids in production by default, which is being documented on the migration guide in https://github.com/web-infra-dev/rspack/pull/13839

Numeric chunk ids cause strict equality (`===`) comparisons in `ChunkExtractor` and `loadableReady` to silently fail against string values. This PR proposes fixing it by coercing all chunk IDs to strings with `String()` before comparison.

## Test plan
Added unit tests for changes, I've also verified manually by applying the diff to my project as well.


panbenson:
@theKashey could you please take a look at this?

tim-soft:
The workaround is to set the rspack config setting `config.optimization.chunkIds: 'named'` which bloats bundles, this would be great to support when you get a moment to review @theKashey

Issue #667, opened 2020-12-07
Implement retry in case of error


ValentinH:
## 🚀 Feature Proposal

Implement a new "retry" option in `@loadable/component`.

## Motivation

We monitor our errors with Sentry and we see quite a lot of errors coming with lazy-loaded components. Indeed, on flaky network, it's easy to have a request aborted and the component is never loaded. In our applications, we have started to wrap all dynamic imports passed to loadable with a retry function such as:
```js
const Example = loadable(() =>
  retry(
    () => import('./Example'),
    { retries: 3 }
  )
)
```

where retry is a simple function:
```ts
type Options = {
  retries?: number
  interval?: number
  exponentialBackoff?: boolean
}

// taken from https://dev.to/goenning/how-to-retry-when-react-lazy-fails-mb5
export function retry<R>(
  fn: () => Promise<R>,
  { retries = 3, interval = 500, exponentialBackoff = true }: Options = {}
) {
  return new Promise<R>((resolve, reject) => {
    fn()
      .then(resolve)
      .catch((error) => {
        setTimeout(() => {
          if (retries === 1) {
            reject(error)
            return
          }

          // Passing on "reject" is the important part
          retry(fn, {
            retries: retries - 1,
            interval: exponentialBackoff ? interval * 2 : interval,
          }).then(resolve, reject)
        }, interval)
      })
  })
}
```

## Example

```js
const Example = loadable(() => import('./Example'),  {retries: 3})
```

## Pitch

I'm wondering if this is something that could be directly implemented as part of `loadable`.

theKashey:
Good point. As a matter of fact - our own Sentry is filled with the same reports.

ValentinH:
If you agree, I'd be happy to work a PR.

theKashey:
That could be a little complicated as right now errors are just thrown, expected to be handled on the nearest ErrorBoundary.
The question - should network problems be still reported (so the error has to be thrown), or in case they are retry-able loadable can swallow them?

ValentinH:
I don't know if you are able to differentiate network errors from other kinds of errors when using dynamic imports. Do you know?

Regarding errors just being thrown at the moment, if the retry is optional, we could still retry all errors as it would not be a breaking change. What do you think?

theKashey:
> I don't know if you are able to differentiate network errors

You can, but that is abstraction leaking, so you should not try to differentiate.

> Regarding errors just being thrown at the moment...

[...]

PR #1037, opened 2026-07-09
feat: add onError and onLoad lifecycle hooks with setConfig

roidayan1:
> Related: #667 — that issue asked for a built-in retry policy (`{ retries }`) and was closed as wontfix. This PR takes the lighter, unopinionated route: a general `onError` hook (mechanism, not policy) that lets consumers implement retry / reload / reporting themselves, plus a symmetric `onLoad`.

## Summary

`loadable` already has an internal `onLoad`, but there is no supported way for a consumer to react to a component load — success or failure. Today a failed dynamic `import()` can only be handled by an Error Boundary, and there is no success callback at all.

This PR adds two small, **opt-in, backward-compatible** lifecycle hooks:

- **`onError(error, props)`** — called when a dynamic import fails. It is a *notification* hook: the error is still thrown afterwards, so your Error Boundary keeps working as the last line of defense. Use it to report the failure or to recover (e.g. reload the page after a deploy invalidated the chunk hashes).
- **`onLoad(result, props)`** — called when a dynamic import resolves.

Both can be set:

```js
// per-instance
loadable(() => import('./X'), {
  onError: (error, props) => reportError(error),
  onLoad:  (result, props) => {},
})

// globally
loadable.setConfig({ onError, onLoad })   // instance option wins over global
```

`lazy` and the `.lib` variants inherit them (same `createLoadable`).

## Why

Reacting to lazy-load failures is a common need (the classic *"Failed to fetch dynamically imported module"* after a deploy). Wrapping every `loadable()` call by hand is repetitive, and there is no global entry point. A first-class hook + a global `setConfig` default covers both, while staying a *mechanism* (recovery policy is left to the consumer) — see docs for a reload-once-with-guard example.

## Notes

- **Backward compatible**: no `onError`/`onLoad` → behavior is byte-for-byte unchanged (error still thrown to the Error Boundary).
- **Fires once per load**, including in suspense mode where the render path can attach several `then`/`catch` handlers to the same cached promise (deduped per instance).
- New tiny modules: `config.js` (global store) and `hooks.js` (instance-over-global resolver). `setConfig` is also exported as a named export.

## Tests & docs

- Tests added to `packages/component/src/loadable.test.js` (`#onError`, `#onLoad`): args, precedence (instance over global), fires-once-in-suspense, and that `onError` still reaches the Error Boundary.
- Docs updated: `api-loadable-component` (options tables + `onError`/`onLoad`/`setConfig` sections) and `error-boundaries` (a stale-chunk reload-once recovery example).


roidayan1:
@theKashey when you have a moment, would you be open to reviewing this? It's small, additive and backward-compatible: two opt-in lifecycle hooks (`onError` / `onLoad`) with per-instance and global (`setConfig`) forms, plus tests and docs. Happy to adjust the API shape or naming to fit your preferences. Thanks!

